PR #53, opened 2020-03-30
Column selector rewrite

johnny-keker:
New ranges selected from context menu multiplies by current selector, unless current column set is a subset of a new one - in that case new selector uses "as is".

### DIFF
--- a/VSRAD.Package/DebugVisualizer/ColumnSelector.cs
+++ b/VSRAD.Package/DebugVisualizer/ColumnSelector.cs
@@ -111,5 +111,16 @@ namespace VSRAD.Package.DebugVisualizer
             foreach (var emptyRegion in emptyRegions)
                 regions.Remove(emptyRegion);
         }
+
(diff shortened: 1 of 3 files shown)


timlathy:
I'd prefer it if you moved the logic from `ColumnSelectorChanged` to a separate method in `ColumnSelector` and added a unit test for it showing both modes of operation (intersection and union).

Also I'm not sure you need `hiddenColumnIndexes`: you could replace `newIndexes.Except(hiddenColumsIndexes)` with something like `newIndexes.Intersect(currentIndexes)`, what do you think?

johnny-keker:
> 
> 
> I'd prefer it if you moved the logic from `ColumnSelectorChanged` to a separate method in `ColumnSelector` and added a unit test for it showing both modes of operation (intersection and union).
> 
> Also I'm not sure you need `hiddenColumnIndexes`: you could replace `newIndexes.Except(hiddenColumsIndexes)` with something like `newIndexes.Intersect(currentIndexes)`, what do you think?

Done.

PR #105, opened 2020-07-08
Function signature did not participate in function detection

Slimakanzer:
## Explanation of the bug
The problem was that the function was not highlighted in the Function List if the caret was on the signature line. 
The function was highlighted only if the cursor was in the scope of the function as in the following pictures:
pr1  pr2

## Solution
This PR solves this problem, now the function is correctly highlighted in the FL. Also now the cursor position is used to highlight the function, not the line.

### DIFF
--- a/VSRAD.Syntax/FunctionList/FunctionList.cs
+++ b/VSRAD.Syntax/FunctionList/FunctionList.cs
@@ -165,7 +165,9 @@ namespace VSRAD.Syntax.FunctionList
         {
             foreach (var func in blocks)
             {
-                if (func.Scope.GetSpan(line.Snapshot).Contains(line.Start))
+                var pointStart = func.TokenStart.GetStart(line.Snapshot);
+                var pointEnd = func.TokenEnd.GetEnd(line.Snapshot);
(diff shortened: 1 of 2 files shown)


VSRAD.Syntax/FunctionList/FunctionList.cs
```diff
@@ -161,15 +160,22 @@ public static void TryUpdateFunctionList(ITextSnapshot version, IReadOnlyList<IB
                 ThreadHelper.JoinableTaskFactory.RunAsync(() => Instance.UpdateFunctionListAsync(version, blocks));
         }
 
-        private static FunctionBlock GetFunctionBy(IEnumerable<FunctionBlock> blocks, ITextSnapshotLine line)
+        private static FunctionBlock GetFunctionBy(IEnumerable<FunctionBlock> blocks, SnapshotPoint position)
```
DrizztDoUrden, line 163: This may be simplified to `=> blocks.FirstOrDefault(func => PointInFunction(func, position));`

Slimakanzer: Got it! Thanks for tips.

PR #106, opened 2020-07-08
"include" keyword supported by asm1 and asm2 parser

Slimakanzer:
**_Include_** keyword  support has been added in this PR.  
External functions, labels and global variables are now defined during parsing of the document and due to this, external tokens can be highlighted and can be used for navigation (GoTo/Peek definition).

The following pictures show the highlight of external functions contained in a _gpr_alloc.s_ file using **_.include "gpr_alloc.s"_** feature.
pr1  pr2


VSRAD.Syntax/Parser/DocumentAnalysisProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+            return CreateDocumentAnalysis(textDocument);
+        }
+
+        private DocumentAnalysis CreateDocumentAnalysis(DocumentInfo document)
+        {
+            var buffer = document.TextBuffer;
+            var asmType = buffer.CurrentSnapshot.GetAsmType();
+
+            DocumentAnalysis documentAnalysis;
+            if (asmType == AsmType.RadAsm2)
```
DrizztDoUrden, line 48: `switch` statement would be preferred for such cases.

VSRAD.Syntax/Parser/Parser.cs
```diff
@@ -50,5 +70,24 @@ public static IBlock SetBlockReady(IBlock block, List<IBlock> list)
...
+                if (documentAnalysis != null)
+                {
+                    foreach (var funcToken in documentAnalysis.LastParserResult.GetGlobalTokens())
+                    {
+                        definitions.Add(new KeyValuePair<AnalysisToken, ITextSnapshot>(funcToken, documentAnalysis.CurrentSnapshot));
+                    }
+                }
+            }
+            catch (ArgumentException)
+            {
```
DrizztDoUrden, line 89: Ditto, the catch should either do something or be nonexistent, possibly replaced by a check.

PR #108, opened 2020-07-10
Slice visualizer

### LINKED ISSUE #71: Feature request: Analyze data
New window specifically dedicated for data analysis across all groups is desired. A very raw mockup:
image

- Status string at the bottom should show info about cell under mouse cursor: Group#, Column#, Value
- Right click context menu should have `Go to watch list` which should open Visualizer window (if not opened), switch to a required `group#`, unhide that value (if hidden), scroll to a corresponding value and select it
- `Show columns` textbox should allow to enter multiple columns like `1-3:5` or `1-3, 5`
- `Watch` combobox should contain only `System` and valid watches
- Each cell should have the same color as a corresponding cell in Visualizer table
- The button icon could be similar to a new Visualizer icon but with cells shaded like a heatmap: image
- Please check that all required data is fetched within single DebugServer command. It seems easy to just always fetch a whole file

johnny-keker:
Addresses #71 

Done:
* Status string at the bottom with info about cell under mouse cursor: Group#, Column#, Value
* Right click context menu has Go to watch list which opens Visualizer window (if not opened), switches to a required group#, unhides that value (if hidden), scrolls to a corresponding value and selects it
* `Show columns` textbox for entering currently displayed columns
* `Watch` combobox contains only `System` and valid watches
* Each cell has the same color as a corresponding cell in Visualizer table
* Slice Visualizer button has an icon similar to Visualizer but with cells shaded like a heatmap
* Limited maximum column count in Slice and Debug Visualizers to 8192 to prevent undefined behavior

Not done yet:
* `Transposed view` feature


VSRAD.Package/DebugVisualizer/SliceVisualizer/SliceCellStyling.cs
```diff
@@ -1,64 +1,116 @@
...
+            {
+                width = _appearance.SliceSubgroupSeparatorWidth;
+                color = _fontAndColor.FontAndColorState.SliceSubgroupSeparatorBrush;
+            }
+            else return;
+
+            // We doing force paint of _visible_ part of cell.
+            // Since we have frozen columns visible part of cell
+            // is not necessarily whole cell.
+            var r = e.CellBounds.Left > _table.RowHeadersWidth
```
timlathy, line 83: What happens if you remove this check? I assume it can only be true for the first displayed column, do we need the `+ 1` offset there? Same goes for `VSRAD.Package.DebugVisualizer.CellStyling`, if we can get rid of the condition on `e.CellBounds.Left`, I think we can share this snippet between the two classes instead of duplicating it.

VSRAD.Package/DebugVisualizer/SliceVisualizer/SliceVisualizerTable.cs
```diff
@@ -26,27 +34,62 @@ public SliceVisualizerTable(IFontAndColorProvider fontAndColor) : base()
...
+            CellMouseEnter += DisplayCellStatus;
+            MouseLeave += (s, e) => _context.StatusString = ""; // clear status bar on leaving control
 
             _state = new TableState(this, 60);
 
             _mouseMoveController = new MouseMove.MouseMoveController(this, _state);
             _selectionController = new SelectionController(this);
             _ = new SliceRowStyling(this);
-            _ = new SliceCellStyling(this, _state, fontAndColor);
+            _ = new SliceCellStyling(this, ColumnStyling, fontAndColor, _context);
```
DrizztDoUrden, line 48: I'd add a static method calling the constructor to these buds and would use it instead of the constructor if you want to keep it like this. It's kind of not a C# way to implement things as side effects of constructor and looks weird.

johnny-keker: Thanks for suggestion, implemented in a9d6a25, did you mean something like this?

VSRAD.Package/DebugVisualizer/SliceVisualizer/SliceRowStyling.cs
```diff
@@ -10,10 +10,16 @@ sealed class SliceRowStyling
         public SliceRowStyling(SliceVisualizerTable table)
         {
             _table = table;
-            _table.RowPostPaint += ReplaceDefaultRowHeaderBitmap;
+            
         }
 
-        private void ReplaceDefaultRowHeaderBitmap(object sender, DataGridViewRowPostPaintEventArgs e)
+        public static void ApplyRowStylingOnRowPostPaint(SliceVisualizerTable table)
```
DrizztDoUrden, line 16: I think you can remove the phrase RowStyling from the method name as it would be used by lines like `SliceRowStyling.ApplyOnRowPostPaint` anyway.
Also, you can make ReplaceDefaultRowHeaderBitmap and constructor private if they are only supposed to be called from within the class.

johnny-keker: Thanks for the suggestion, resolved in addbe4a

VSRAD.Package/DebugVisualizer/GroupIndexSelector.cs
```diff
@@ -151,6 +151,8 @@ public void Update()
             ErrorsChanged?.Invoke(this, new DataErrorsChangedEventArgs(nameof(Z)));
         }
 
+        public void SetGroupIndex(int index) => X = (uint)index;
```
timlathy, line 154: I think this method can be safely removed and all usages replaced with `GoToGroup`, which (in theory, I haven't done much testing) should also work when 3D NDRange is enabled.

VSRAD.Package/DebugVisualizer/MouseMove/MouseMoveController.cs
```diff
@@ -20,7 +20,7 @@ public MouseMoveController(DataGridView table, TableState state)
             {
                 new ScaleOperation(_table, _state),
                 new PanOperation(_state),
-                new ReorderOperation(_table)
+                new ReorderOperation(_table, _state.ReorderingEnabled)
```
timlathy, line 23: I think you can just add an `if` statement here:
```
_operations = new List<IMouseMoveOperation>
{
    new ScaleOperation(_table, _state),
    new PanOperation(_state)
};
if (_state.ReorderingEnabled)
  _operations.Add(new ReorderOperation(_table));
```

This way, you don't need to change anything in the `ReorderOperation` class.

VSRAD.Package/DebugVisualizer/SliceVisualizer/SliceVisualizerContext.cs
```diff
@@ -48,6 +48,48 @@ public SliceVisualizerContext(Options.ProjectOptions options, VisualizerContext
...
+            }
+
+            var gNum = SelectedWatchView.GetGroupIndex(row, column);
+            var lNum = SelectedWatchView.GetLaneIndex(column);
+            StatusString = $"Group# {gNum}, Column# {lNum}, Value: {val}";
+        }
+
+        public void NavigateToCell(int sliceRowIndex, int sliceColumnIndex)
+        {
+            Options.VisualizerOptions.NDRange3D = false; // disable NDRange3D since we dont have implementation of navigation for it
```
timlathy, line 72: See my suggestion about replacing `SetGroupIndex` with `GoToGroup`, which doesn't require this hack.

timlathy (changes requested):

Slimakanzer (changes requested):
The slice visualizer works well, but sometimes debug extension shows *OutOfMemory* and *Arithmetic* exception messages when changing the *Show columns* field. For example ***0:1-128*** value.

VSRAD.Package/DebugVisualizer/SliceVisualizer/TypedSliceWatchView.cs
```diff
@@ -124,14 +125,14 @@ private static void GetMinMaxValues(SliceWatchView view, VariableType type, out
...
-                            if (!float.IsNaN(firstHalf))
+                            if (!float.IsNaN(firstHalf) && !float.IsInfinity(firstHalf))
                             {
                                 if (firstHalf < fmin)
                                     fmin = firstHalf;
                                 if (firstHalf > fmax)
                                     fmax = firstHalf;
                             }
-                            if (!float.IsNaN(secondHalf))
+                            if (!float.IsNaN(secondHalf) && !float.IsInfinity(firstHalf))
```
Kirpich30000, line 135: should probably be `IsInfinity(secondHalf)`

johnny-keker: Thanks, fixed in 43da773

johnny-keker:
> 
> 
> The slice visualizer works well, but sometimes debug extension shows _OutOfMemory_ and _Arithmetic_ exception messages when changing the _Show columns_ field. For example _**0:1-128**_ value.

Fixed in 6e425bb

PR #110, opened 2020-07-13
Profile redesign

### LINKED ISSUE #90: Debug profiles redesign
Redesign is needed to support additional usecases and make configuration more flexible. A draft overview:
 We need to introduce basic actions like `execute local`, `execute remote`, `copy to remote`, `copy to local`, `open in VS`. Each action will have it's own settings like paths, arguments, etc depending on action type. Then any RAD command would be just a sequence of basic actions with some additional properties depending on command type (watch file for debug, etc.)

timlathy:
Resolves #90 

This PR replaces hardcoded tasks (profile, preprocess, disassemble) with a system of user-defined actions. Actions are composed of four basic steps: `Execute` external command, `Copy File`, `Open in Editor`, `Run Action`.

Actions are executed by `VSRAD.Package.Server.ActionRunner`, which produces a `VSRAD.Package.Server.ActionRunResult` containing error info, log messages and time spent. It is displayed in the output window by `VSRAD.Package.ProjectSystem.ActionLogger`.

`Debug` action is predefined in every profile. In addition to steps, it includes configuration for debugger-specific output/status/watch files.

Hardcoded macros ("$(RadDebugOutputPath)") are replaced with user-defined macros, except for values that are set in the visualizer window ("$(RadWatches)", "$(RadCounter)", etc.) — they are unchanged. 

Profiles created in older versions are automatically converted to the new system. The importer transforms configuration for hardcoded tasks into identical actions. For instance, `Disassembler` becomes a sequence of 1. `Execute`, 2. `Copy File`, 3. `Open in Editor` steps.

Known issues (will be addressed in subsequent PRs):
• Local debugger execution is not supported. (It requires reworking the current logic for fetching output file in groups.) If the output file location is set to `Local`, a descriptive error message is shown.
• Minor UI problems in the redesigned profile editor. Notably, the scroll may jump to the top of the step list when selecting a step.
• Most of the obsolete (and now unused) code hasn't been removed yet.


VSRAD.DebugServer/IPC/Commands.cs
```diff
@@ -13,8 +16,9 @@ public static ICommand ReadCommand(this IPCReader reader)
                 case 0: return Execute.Deserialize(reader);
                 case 1: return FetchMetadata.Deserialize(reader);
                 case 2: return FetchResultRange.Deserialize(reader);
-                case 3: return Deploy.Deserialize(reader);
-                case 4: return ListEnvironmentVariables.Deserialize(reader);
+                case 3: return PutFileCommand.Deserialize(reader);
+                case 4: return Deploy.Deserialize(reader);
+                case 5: return ListEnvironmentVariables.Deserialize(reader);
```
Slimakanzer, line 21: It would be better to define byte type enum with command types `enum CommandType : byte { Execute = 0, ..., ListEnvironmentVariables = 5 }`

VSRAD.DebugServer/IPC/Commands.cs
```diff
@@ -27,8 +31,9 @@ public static void WriteCommand(this IPCWriter writer, ICommand command)
...
                 case FetchMetadata _: commandType = 1; break;
                 case FetchResultRange _: commandType = 2; break;
-                case Deploy _: commandType = 3; break;
-                case ListEnvironmentVariables _: commandType = 4; break;
+                case PutFileCommand _: commandType = 3; break;
+                case Deploy _: commandType = 4; break;
+                case ListEnvironmentVariables _: commandType = 5; break;
                 default: throw new ArgumentException($"Unable to serialize {command.GetType()}");
             }
             writer.Write(commandType);
```
Slimakanzer, line 39: It’s also better to use enum value assignment (in combination with my previous suggestion) and then cast `commandType` variable to `byte` type

VSRAD.DebugServer/IPC/Responses.cs
```diff
@@ -14,7 +14,8 @@ public static IResponse ReadResponse(this IPCReader reader)
                 case 0: return ExecutionCompleted.Deserialize(reader);
                 case 1: return MetadataFetched.Deserialize(reader);
                 case 2: return ResultRangeFetched.Deserialize(reader);
-                case 3: return EnvironmentVariablesListed.Deserialize(reader);
+                case 3: return PutFileResponse.Deserialize(reader);
+                case 4: return EnvironmentVariablesListed.Deserialize(reader);
```
Slimakanzer, line 18: Here, by the same principle as in my previous suggestion

VSRAD.Package/Commands/QuickActionCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+                if (actionError is Error e1)
+                    Errors.Show(e1);
+            }
+            finally
+            {
+                await _statusBar.ClearAsync();
+            }
+        }
+    }
+}
```
Slimakanzer, line 99: Here is the repetition of the same function as `ActionsMenuCommand`. It might be better to add some abstraction for `ICommandHandler` that works with `VsStatusBarWriter`

timlathy: Addressed in db4cd27 and 04edcbd

VSRAD.DebugServer/Handlers/PutFileHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+        public PutFileHandler(PutFileCommand command)
+        {
+            _command = command;
+        }
+
+        public Task<IResponse> RunAsync()
+        {
+            var fullPath = Path.Combine(_command.WorkDir, _command.Path);
+            var status = DoPutFile(fullPath, _command.Data);
+            return Task.FromResult<IResponse>(new PutFileResponse { Status = status });
```
DrizztDoUrden, line 22: This method is not really async. In fact, it just creates a completes task with a result. You may want to use something like this: [TaskFactory.StartNew](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskfactory.startnew?view=netcore-3.1) If you want to start some actual work in a new task.

VSRAD.DebugServer/ObservableProcess.cs
```diff
@@ -0,0 +1,162 @@
...
+            {
+                StartInfo = new ProcessStartInfo(command.Executable, command.Arguments)
+                {
+                    WorkingDirectory = command.WorkingDirectory
+                },
+                EnableRaisingEvents = true
+            };
+            if (command.RunAsAdministrator)
+            {
+                _process.StartInfo.UseShellExecute = true;
```
DrizztDoUrden, line 32: Are you sure you don't need to redirect output here? UseShellExecute with the value of true redirects it to the console.

VSRAD.DebugServer/ObservableProcess.cs
```diff
@@ -0,0 +1,162 @@
...
+            {
+                return new ExecutionCompleted { Status = ExecutionStatus.CouldNotLaunch };
+            }
+
+            await processExitedTcs.Task;
+
+            var status = _stoppedByTimeout ? ExecutionStatus.TimedOut : ExecutionStatus.Completed;
+
+            var stdout = await stdoutTask;
+            var stderr = await stderrTask;
```
DrizztDoUrden, line 81: I don't see where you finish this tasks. Are you sure you need the tasks here at all? You already have a tcs you create checking on process exit. I guess you don't need tasks for stdout/stderr at all here, just stringbuilders.

VSRAD.Package/Commands/EvaluateSelectedCommand.cs
```diff
@@ -19,10 +19,12 @@ sealed class EvaluateSelectedCommand : BaseRemoteCommand
         private readonly IProject _project;
         private readonly IActiveCodeEditor _codeEditor;
         private readonly QuickInfoEvaluateSelectedState _quickInfoState;
+#pragma warning disable IDE0052 // Remove unread private members
         private readonly IFileSynchronizationManager _deployManager;
         private readonly IOutputWindowManager _outputWindow;
         private readonly ICommunicationChannel _channel;
         private readonly IErrorListManager _errorListManager;
```
DrizztDoUrden, line 26: Do you really need unread fields here?

VSRAD.DebugServer/ObservableProcess.cs
timlathy: The tasks are completed when `OutputDataReceived` and `ErrorDataReceived` events fire with `null`, which indicates
that there's no more output. This can happen after the `Exited` event due to output buffering.

If I understood your suggestion correctly, we should return `StringBuilder`s from `InitializeOutputCapture`
and read them after `await processExitedTcs.Task;`. This may lead to a loss of data for large outputs, as demonstrated by [this patch](https://gist.github.com/timlathy/7eda75dd19d547e0c140f80d4ea94440) (the test fails with modifications and passes without them, i.e. with `await`).

VSRAD.DebugServer/ObservableProcess.cs
timlathy: You're right, we don't currently capture the output when running commands as administrator because we launch them via cmd + `runas`, which takes care of the UAC prompt for us.

The problem is, the standard `Process` API (`RedirectStandardOutput`) cannot be used together with `UseShellExecute`
("The Process object must have the UseShellExecute property set to false in order to redirect IO streams.").

I'm sure we can find an alternative approach for running a process with eleavated privileges while redirecting its IO streams, but I'm afraid it will require using native APIs and, consequently, much more code. This is something that we probably should file a separate issue for.

VSRAD.DebugServer/Handlers/PutFileHandler.cs
timlathy: Fixed it, thanks!

VSRAD.Package/Commands/EvaluateSelectedCommand.cs
timlathy: `EvaluateSelectedCommand` is one of the many obsolete parts of our codebase. It is not used anymore, the whole class should just be deleted, along with related components in `VSRAD.Package.ProjectSystem.EditorExtensions`.

Cleaning it up is in our backlog, but it's not a part of this PR, since as you correctly noted it's already way too big. The changes that were made are necessary to temporarily suppress a compile warning and remove dependencies on parts that were changed.

VSRAD.DebugServer/IPC/Commands.cs
timlathy: That's a great suggestion, thanks! I'll apply it in subsequent PRs along with command/response internal name changes as indicated by this TODO
```
// TODO: Rename commands and responses to follow a predictable pattern of XCommand -> XResponse
// (e.g. ExecuteCommand and ExecuteResponse instead of Execute and ExecutionFinished)
```

PR #111, opened 2020-07-20
Profile editor bug fixes

timlathy:
Macro list:
* Don't use expanders to display predefined/user-defined macro groups (d94b4b7)
* A single empty macro row is displayed at a time, i.e. clicking the "+" button repeatedly does not result in multiple empty rows (6c0bfe0, bf88153)
* Selecting multiple macros and pressing Delete only removes macros that were selected (ab776c4, caused by `GetHashCode` referencing mutable fields and `DataGrid` using object hashes to track selected items)
* Macro editor no longer hangs when a profile contains recursive macros (2f236b3, includes a unit test that reproduces the problem)

Toolbar:
* Add an option to remove toolbar action mapping (35f0e05)
* Show action name in toolbar button tooltip (862a61f)
* Display toolbar button icons in the "Toolbar" editor section (6dc4112)

Profile editor:
* Run Action step dropdown now shows all available actions (f16b742)
* A confirmation dialog is shown before deleting macros (ab776c4), actions and profiles (686d274)
* Profile and action names are validated to be non-empty (7402e4d)
* Removing a profile no longer causes a crash (47a8d71, includes a unit test that reproduces the problem)

Theming:
* Changed edit and move up/move down icons so they are properly displayed on Windows 7 (eb5e8ce)
* Fixed dark-on-dark text in action editor context menu on Windows 7 (f3bcc36)

A follow-up PR will include additional improvements (directory copy step, active source file macro changes, evaluating macros in step description, etc.), this one is focused on bug fixes and small tweaks.


timlathy:
> VS crashes if the value of the _Test2_ macro is changed to _$(Test2)_ using the attached [conf.json](https://github.com/vsrad/radeon-asm-tools/files/4948726/RADProject1.radproj.conf.zip)

Thanks! Fixed in cd536d0

PR #112, opened 2020-07-20
Syntax bug fixes

Slimakanzer:
Parser:
 * Fixed a critical error if the shader source code was written incorrectly (function parameters have the same names), code highlighting and FL did not work (https://github.com/vsrad/radeon-asm-tools/commit/aab349391c3dc04c3aeecbc071600470c305894e)
 * Fixed error when opening an already open source file. This can be reproduced when opening a file using RadeonAsmDebugger (https://github.com/vsrad/radeon-asm-tools/pull/112/commits/f02e2316341a5337ef24926a8da37e2bc7c83741)

Function list:
 * Fixed filtering of the FL when changing the code window (if the filter was set). The filter was not applied before (https://github.com/vsrad/radeon-asm-tools/commit/971e212f61e860b7c3575dd3fa1ccb1b0b90a6be)
 * Fixed `Line Number` column width with large numbers (https://github.com/vsrad/radeon-asm-tools/pull/112/commits/9a718d8eaebad44cf34f6f929206b95aa4f13aeb)

Other:
 *  Fixed collapse when opening a file for the first time, previously it was applied only after changing the file (https://github.com/vsrad/radeon-asm-tools/commit/63e6f5fd97bed5f612ccae069dd2d6a76a3a0685)
 * Fixed display of tokens in quick info tooltip, now they are grouped by filepaths (if there is more than one). It is especially noticeable if several declarations of one instruction in the instruction documentation (https://github.com/vsrad/radeon-asm-tools/commit/ee80ced0a37c0e6dba8b595a6f6ade6326b7d64c)

### DIFF
--- a/VSRAD.Syntax/Collapse/OutliningTagger.cs
+++ b/VSRAD.Syntax/Collapse/OutliningTagger.cs
@@ -19,6 +19,7 @@ namespace VSRAD.Syntax.Collapse
         public OutliningTagger(DocumentAnalysis documentAnalysis)
         {
             currentSpans = new List<Span>();
+            ParserUpdated(documentAnalysis.CurrentSnapshot, documentAnalysis.LastParserResult);
             documentAnalysis.ParserUpdated += ParserUpdated;
         }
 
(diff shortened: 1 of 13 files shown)


Slimakanzer:
> Done some testing, looks good, however I noticed little flaws:
> 
> * Sometimes Function List's `Line` column don't display last valuable digit (line 1331 -> 133). I would suggest to set `Line` column min width to fit at least 5 digits to prevent that behaviour
> * Function List can display wrong macro location when the cursor is traversing through the nested macro declarations from bottom to the top

Thanks!
 * FL Line number fixed in (https://github.com/vsrad/radeon-asm-tools/pull/112/commits/9a718d8eaebad44cf34f6f929206b95aa4f13aeb)
 * `Current function` with nested macro declaration fixed in  (https://github.com/vsrad/radeon-asm-tools/pull/112/commits/5d3b1e0123e6a4c4ef7f1cf05a35fad0a80ff4c9)

PR #115, opened 2020-07-23
Unevaluated watches handling fixes

johnny-keker:
* Fixed unevaluated watches painting
* Gray out all watches before debug session start
* Properly gray out watches if debug session was unsuccessfull

### DIFF
--- a/VSRAD.Package/DebugVisualizer/VisualizerTable.cs
+++ b/VSRAD.Package/DebugVisualizer/VisualizerTable.cs
@@ -308,7 +308,7 @@ namespace VSRAD.Package.DebugVisualizer
 
         #region Styling
 
-        public void GrayOutColumns()
+        public void GrayOutRows()
         {
             if (ShowSystemRow)
                 RowStyling.GrayOutRow(_fontAndColor.FontAndColorState, Rows[0]);
@@ -362,7 +362,6 @@ namespace VSRAD.Package.DebugVisualizer
                 _fontAndColor.FontAndColorState);
             columnStyling.Apply(_state.DataColumns);
 
-            
             foreach (DataGridViewRow row in Rows)
                 RowStyling.UpdateRowHighlight(row, _fontAndColor.FontAndColorState, _getValidWatches());
 
--- a/VSRAD.Package/DebugVisualizer/VisualizerTable.cs
+++ b/VSRAD.Package/DebugVisualizer/VisualizerTable.cs
@@ -309,7 +309,7 @@ namespace VSRAD.Package.DebugVisualizer
         #region Styling
 
         public void GrayOutColumns()
-        {;
+        {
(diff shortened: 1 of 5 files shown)


VSRAD.Package/DebugVisualizer/VisualizerTable.cs
```diff
@@ -302,19 +308,29 @@ private static void LockWatchRowForEditing(DataGridViewRow row, bool canBeRemove
 
         #region Styling
 
-        public void GrayOutColumns(uint groupSize)
+        public void GrayOutColumns()
```
timlathy, line 311: The name should be changed too (`GrayOutRows`?)

VSRAD.Package/DebugVisualizer/VisualizerTable.cs
```diff
@@ -346,8 +362,9 @@ public void ApplyDataStyling(ProjectOptions options, uint groupSize, Server.Watc
                 _fontAndColor.FontAndColorState);
             columnStyling.Apply(_state.DataColumns);
 
+
```
timlathy, line 365: Formatting

VSRAD.Package/DebugVisualizer/VisualizerTable.cs
```diff
@@ -302,19 +308,29 @@ private static void LockWatchRowForEditing(DataGridViewRow row, bool canBeRemove
 
         #region Styling
 
-        public void GrayOutColumns(uint groupSize)
+        public void GrayOutColumns()
         {
-            _computedStyling.GrayOutColumns(groupSize);
```
timlathy, line 307: This looks like the only usage of `ComputedColumnStyling.GrayOutColumns`, you can remove it. I don't think `ColumnStyling.GrayOutColumns` is used anywhere either, that can be removed too.

VSRAD.Package/DebugVisualizer/VisualizerTable.cs
johnny-keker: Thanks! Resolved in c1dce41

VSRAD.Package/DebugVisualizer/VisualizerTable.cs
johnny-keker: Thanks! Resolved in c1dce41

VSRAD.Package/DebugVisualizer/VisualizerTable.cs
johnny-keker: Thanks! Resolved in c1dce41

johnny-keker:
> 
> 
> The last line is initially white, but if you change the group size, it will be colored gray. If you add a new watch after that, then it will be colored white again.

Thanks, fixed in d71e1d2

PR #119, opened 2020-07-25
Bug fixes

johnny-keker:
Fixed:

* Visualizer table `Delete` key handling
* Unwanted auto setting of toolbar commands actions
* `Illegal characters in path` not descriptive error message
* Whitespaces in action properties


johnny-keker:
@timlathy  and @Slimakanzer  thanks for suggestions!

* _Local to Remote_ situation handled in 62bdb6b
* Tests added in b507983

Looking forward for new reviews and approval

PR #123, opened 2020-07-29
Rad group size

johnny-keker:
Added `RadGroupSize` macro that represents group size from visualizer table interface

### DIFF
--- a/VSRAD.Package/Options/DebuggerOptions.cs
+++ b/VSRAD.Package/Options/DebuggerOptions.cs
@@ -39,7 +39,7 @@ namespace VSRAD.Package.Options
         public bool SingleActiveBreakpoint { get => _singleActiveBreakpoint; set => SetField(ref _singleActiveBreakpoint, value); }
 
         private uint _groupSize = 512;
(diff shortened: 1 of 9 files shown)


VSRAD.Package/Options/DebuggerOptions.cs
```diff
@@ -38,6 +38,9 @@ public sealed class DebuggerOptions : DefaultNotifyPropertyChanged
         private bool _singleActiveBreakpoint = false;
         public bool SingleActiveBreakpoint { get => _singleActiveBreakpoint; set => SetField(ref _singleActiveBreakpoint, value); }
 
+        private uint _groupSize = 512;
+        public uint GroupSize { get => _groupSize; set => SetField(ref _groupSize, value == 0 ? 512 : value); }
```
Slimakanzer, line 42: The assigned default value of `private` member is meaningless because it is verified in SetFeld.
```suggestion
        private uint _groupSize;
        public uint GroupSize { get => _groupSize; set => SetField(ref _groupSize, value == 0 ? 512 /* default value */ : value); }
```

PR #145, opened 2020-08-24
Tie props to solution

### LINKED ISSUE #142: Profile is lost when switching between projects
Profile and debugger configuration is lost when switching(`Set as Startup Project`) between projects within a solution.

### LINKED ISSUE #143: Tie debugger configuration to a solution instead of a project
Is it possible to tie json configuration to a solution instead of a project?
If so we will need to evaluate $() vars differently based on current active project.

johnny-keker:
Addresses #142 
Addresses #143 

From now config file is assigned to the whole solution, not the project. That means that our configuration will be stable when working with the solutions that contain several projects

### DIFF
--- a/VSRAD.Package/ProjectSystem/SolutionProperties.cs
+++ b/VSRAD.Package/ProjectSystem/SolutionProperties.cs
@@ -21,12 +21,15 @@ namespace VSRAD.Package.ProjectSystem
             var solutionPath = solution.FullName;
             var legacyOptionsPath = project.FullPath + ".conf.json";
             var legacyOptionsPathAternative = project.FullPath + ".user.json";
+            var solutionConfigPath = !string.IsNullOrWhiteSpace(solutionPath) ? GetConfigPath(solutionPath) : "";
 
-            Options = File.Exists(legacyOptionsPath)
-                ? ProjectOptions.Read(legacyOptionsPath)
-                : File.Exists(legacyOptionsPathAternative)
-                    ? ProjectOptions.ReadLegacy(legacyOptionsPathAternative)
-                    : new ProjectOptions();
+            Options = !string.IsNullOrWhiteSpace(solutionPath) && File.Exists(solutionConfigPath)
+                ? ProjectOptions.Read(solutionConfigPath)
+                : File.Exists(legacyOptionsPath)
+                    ? ProjectOptions.Read(legacyOptionsPath)
+                    : File.Exists(legacyOptionsPathAternative)
+                        ? ProjectOptions.ReadLegacy(legacyOptionsPathAternative)
+                        : new ProjectOptions();
 
             _optionsFilePath = string.IsNullOrWhiteSpace(solutionPath)
                 // we opened project outside of any solution
@@ -34,7 +37,7 @@ namespace VSRAD.Package.ProjectSystem
                 ? legacyOptionsPath
                 // we opened solution with at least one project
                 // we'll save profiles to the new location in solution dir
-                : GetConfigPath(solutionPath);
+                : solutionConfigPath;
 
             Options.PropertyChanged += OptionsPropertyChanged;
             Options.DebuggerOptions.PropertyChanged += OptionsPropertyChanged;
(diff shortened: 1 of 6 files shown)


VSRAD.Package/ProjectSystem/SolutionProperties.cs
```diff
@@ -0,0 +1,71 @@
...
+            // we'll save profiles to the new location in solution dir
+            else
+            {
+                Options = File.Exists(legacyOptionsPath)
+                    ? ProjectOptions.Read(legacyOptionsPath)
+                    : File.Exists(legacyOptionsPathAternative)
+                        ? ProjectOptions.ReadLegacy(legacyOptionsPathAternative)
+                        : new ProjectOptions();
+                _optionsFilePath = GetConfigPath(solutionPath);
+            }
```
Slimakanzer, line 50:
```suggestion
            Options = File.Exists(legacyOptionsPath)
                ? ProjectOptions.Read(legacyOptionsPath)
                : File.Exists(legacyOptionsPathAternative)
                    ? ProjectOptions.ReadLegacy(legacyOptionsPathAternative)
                    : new ProjectOptions();
            _optionsFilePath =  string.IsNullOrWhiteSpace(solutionPath)
                // we opened project outside of any solution
                // load legacy path if exist, load legacy alternative if exists
                ? legacyOptionsPath
                // we opened solution with at least one project
                // trying to retrieve any version of obsolete profiles
                : GetConfigPath(solutionPath);
```

johnny-keker: Yep, thanks for suggestion, resolved in dcd969a

Slimakanzer (changes requested):
The current implementation never tries to read options from the `json` file, which may be in the solution directory. 
For example:
* Create new project
* The extension will try to find options in the project directory (`legacyOptionsPath ` and `legacyOptionsPathAternative `)
* The extension won't find files and will create new options
* Restart project and extension will try to find options in the project directory again
* The extension won't find files again and **will create new options**

johnny-keker:
> 
> 
> The current implementation never tries to read options from the `json` file, which may be located in the solution directory.
> For example:
> 
>     * Create new project
> 
>     * The extension will try to find options in the project directory (`legacyOptionsPath ` and `legacyOptionsPathAternative `)
> 
>     * The extension won't find files and will create new options
> 
>     * Restart project and extension will try to find options in the project directory again
> 
>     * The extension won't find files again and **will create new options**

Thanks! That's totally true, fixed in 2d730ee

VSRAD.Package/ProjectSystem/SolutionProperties.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            Options = !string.IsNullOrWhiteSpace(solutionPath) && File.Exists(solutionConfigPath)
+                ? ProjectOptions.Read(solutionConfigPath)
+                : File.Exists(legacyOptionsPath)
+                    ? ProjectOptions.Read(legacyOptionsPath)
+                    : File.Exists(legacyOptionsPathAternative)
+                        ? ProjectOptions.ReadLegacy(legacyOptionsPathAternative)
+                        : new ProjectOptions();
+
+            _optionsFilePath = string.IsNullOrWhiteSpace(solutionPath)
```
Slimakanzer, line 34: Here is a repetition of `string.IsNullOrWhiteSpace(solutionPath)`. It would be better to add local variable `var isSolutionExists = string.IsNullOrWhiteSpace(solutionPath);` which also improves understanding of the code.

Slimakanzer (changes requested):

VSRAD.Package/ProjectSystem/SolutionProperties.cs
johnny-keker: Yep, resolved in eff6af0

Slimakanzer (commented):
LGTM, works as expected
* Tried open old projects
* Tried open new projects
* Tried add second project
* Tried change active project
* Tried open project without solution

Slimakanzer (approved):

johnny-keker:
Superseded by #158

Issue #147, opened 2020-08-25
Error while showing project debug properties


Kirpich30000:
Switching to `Configuration Properties`->`Debugging` causes following error:
![image](https://user-images.githubusercontent.com/12259057/91222627-21343880-e728-11ea-99d8-c6432869d279.png)

Which is basically a show stopper for using native C debugger and it affects all projects.

PR #148, opened 2020-08-26
fixed no data source in this context error

johnny-keker:
Addresses #147 

Fixes "No data source in this context" error when using RAD debugger in VisualC projects.

Note that you'll have to copy new version of `RADDebugger.xaml` to the `ImportAfter/` directory to apply the fix


Slimakanzer:
LGTM

* Tried to reproduce error with 16.5.5 patch using `develop` version (no error appeared)
* Tried to reproduce error with 16.7.2 patch using `develop` version (error appeared)
* Tried to reproduce error with 16.7.2 patch using this branch version (no error appeared)

Issue #133, opened 2020-08-05
Watch list rearrange issues


Kirpich30000:
Moving watches around with mouse doesn't always works properly, especially when trying to move last watch.

Issue #130, opened 2020-08-05
Omit empty brackets when adding new watch


Kirpich30000:
The "Add to Watches" should remove empty brackets but still preserve non-empty. For example:

- "acc[]" -> "acc" should be added to a watch list
- "acc[i]" -> "acc[i]" should be added to a watch list

"Add to Watches" behaviour when text is selected is unchanged - add selected text exactly.

PR #149, opened 2020-08-27
Visualizer fixes

johnny-keker:
Addresses #130 
Addresses #133 

* Fixed reordering issue - after adding a new watch it was impossible to change its position
* Removing empty brackets from watch name

### DIFF
--- a/VSRAD.Package/Commands/AddToWatchesCommand.cs
+++ b/VSRAD.Package/Commands/AddToWatchesCommand.cs
@@ -2,6 +2,7 @@
 using Microsoft.VisualStudio.ProjectSystem;
 using System;
 using System.ComponentModel.Composition;
+using System.Text.RegularExpressions;
 using VSRAD.Package.ProjectSystem;
 using VSRAD.Package.ToolWindows;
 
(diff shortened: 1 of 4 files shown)


Slimakanzer (approved):
LGTM, works as expected!
* Tried to use `AddToWatches` with empty square brackets
* Tried to use `AddToWatches` with not empty square brackets
* Tried reorder watches in different situations like after adding, after deleting, after successful debug result, after invalid debug result and etc.

Slimakanzer (changes requested):
It would be nice to implement removing empty brackets in the `ActiveCodeEditor.GetActiveWord()`. In this case, you will not need to use an dte instance, switch to main thread and check selected text again.

johnny-keker:
> 
> 
> It would be nice to implement removing empty brackets in the `ActiveCodeEditor.GetActiveWord()`. In this case, you will not need to use an dte instance, switch to main thread and check selected text again.

Thanks for the suggestion, implemented in 4b7335b

Slimakanzer (approved):
LGTM

Issue #150, opened 2020-08-27
No active projects found message when deleting a project


Slimakanzer:
The message `No active projects found` is displayed in the RAD Options window when deleting/unloading a project even if there are more RadeonAsm projects. After that, a restart of the Visual Studio is required.

PR #153, opened 2020-08-28
Handle project deletion

johnny-keker:
Addresses #150 

Prevents the UI blocking with message `No active projects found` after deleting of the project if it was not the last one in solution


Slimakanzer (approved):
LGTM, works as expected
* Tried to delete projects in the solution
* Tried to delete all projects and then add RadAsm project
* Tried to delete project which is not in solution

timlathy (changes requested):
Superseded by #158

PR #156, opened 2020-08-31
Prevent configuration file corruption

### LINKED ISSUE #154: JSON gets corrupted in case of hang/hard reset
JSON config file may become NULNULNULNULNULNULNUL … sequence if the host hangs and the system needs to be HW ressetted.

timlathy:
Resolves #154 

Currently, project configuration is saved using `File.WriteAllText`, which does not flush file contents to disk, so a system hang/crash can leave the file in a corrupted state (filled with zero bytes).

To ensure that the configuration file always remains in a valid state, we first create a temporary file (`.conf.json.tmp`), then rename it to `.conf.json` after the data is written to disk.

### DIFF
--- a/VSRAD.Package/Options/ProjectOptions.cs
+++ b/VSRAD.Package/Options/ProjectOptions.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Windows.Forms;
 using VSRAD.Package.ProjectSystem.Profiles;
 using VSRAD.Package.Utils;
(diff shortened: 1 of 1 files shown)


johnny-keker (approved):
Done some testing:

* Edited the project config
* Made project directory read-only and edited the project config
* Made config file read-only and edited the project config

It the last two cases I was asked whether extension should make config file writable and in both cases I was able to successfully save new config

However, I couldn't reproduce the bug from #154 since it require me to crash my system in exact moment when extension is trying to write new config to file, that seems to be impossible

Aside from that, LGTM, works as expected

PR #157, opened 2020-08-31
Fix crash when configuration file is empty

### LINKED ISSUE #155: VS stops working if JSON config is corrupted
If the content of *.conf.json disappeared (become NULNULNULNULNULNULNUL … sequence), the VS IDE very much stops working properly. Not only it does not see/does not allow the RADDebugger anymore, but it cannot access any standard project properties on the C++ side (not only the Debug ones), nor it is possible to run  the C++ debugger.

timlathy:
Resolves #155


johnny-keker (approved):
Done some testing:

* Tried to open a project with empty config on `dev` branch, got a crash
* Tried to do the same on `config-read-null-fix`, got new config

LGTM, works as expected

PR #158, opened 2020-09-02
Load configuration from current startup project

### LINKED ISSUE #142: Profile is lost when switching between projects

timlathy:
Resolves #142 and #150.

This PR adds a `SolutionManager` class that observes startup project changes and loads the appropriate configuration file. It replaces `ProjectLifecycle`, which was never intended to support multiple RAD projects in a single solution.


johnny-keker (approved):
Done some testing:

* Created new RAD-project, configured and run debug
* Added new RAD-project to solution, configured its debug, set first project as startup and run the debug, then new project again and run the debug
* Same as previous but with VisualC project
* Added existed RAD-project and repeated test routine
* Same as previous but with VisualC project

LGTM, works as expected

PR #161, opened 2020-09-04
Fix race condition when sending multiple commands in parallel

johnny-keker:
Sending multiple commands in parallel could lead to unexpected behavior

### DIFF
--- a/VSRAD.Package/Commands/ActionsMenuCommand.cs
+++ b/VSRAD.Package/Commands/ActionsMenuCommand.cs
@@ -23,7 +23,8 @@ namespace VSRAD.Package.Commands
         private readonly SVsServiceProvider _serviceProvider;
         private readonly VsStatusBarWriter _statusBar;
 
-        private bool _actionRuns;
+        private bool _actionRuns = false;
+        private string _currentActionName;
         private ProfileOptions SelectedProfile => _project.Options.Profile;
 
         [ImportingConstructor]
@@ -40,7 +41,6 @@ namespace VSRAD.Package.Commands
             _serviceProvider = serviceProvider;
             _deployManager = deployManager;
             _statusBar = new VsStatusBarWriter(serviceProvider);
-            _actionRuns = false;
         }
 
         public Guid CommandSet => Constants.ActionsMenuCommandSet;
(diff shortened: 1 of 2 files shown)


VSRAD.Package/Commands/ActionsMenuCommand.cs
```diff
@@ -110,19 +112,28 @@ private async Task ExecuteActionAsync(ActionProfileOptions action)
         {
             try
             {
+                if (_actionRuns)
+                {
+                    Errors.Show(new Error($"Action {action.Name} is running.\nIf you think that this may be hang, please use Disconnect button."));
```
timlathy, line 117: There's no need for extra boxing, `Errors.ShowWarning(string)` is identical in this context.

Also I'd slightly alter the message to include `_currentActionName` instead of `action.Name`, like:
 "Action {_currentActionName} is already running.\nIf you believe this to be a hang, use the Disconnect button available in Tools -> RAD Debug -> Options to abort the currently running action."

VSRAD.Package/Commands/ActionsMenuCommand.cs
```diff
@@ -133,6 +144,11 @@ private async Task ExecuteActionAsync(ActionProfileOptions action)
...
                 if (actionError is Error runError)
                     Errors.Show(runError);
+                _actionRuns = false;
+            }
+            catch (OperationCanceledException)
+            {
+                _actionRuns = false;
             }
             finally
             {
```
timlathy, line 154: I suggest you to move `_actionRuns = false;` to the `finally` block. This way, it is always executed when leaving the function and you don't need to repeat it multiple times.

timlathy (changes requested):

VSRAD.Package/Commands/ActionsMenuCommand.cs
johnny-keker: Thanks for the suggestion!
Resolved by 380d70c, c591575 and 1d60def.

VSRAD.Package/Commands/ActionsMenuCommand.cs
johnny-keker: Thanks! Resolved by 1d60def.

timlathy (approved):

PR #166, opened 2020-09-23
Filter by type button in the function list

Slimakanzer:
Added filter by type of tokens in the FL. 
Filter has three states:
 * FL - functions and labels are included in the list
 * F - only functions are included in the list
 * L - only labels are included in the list

These states change cyclically when you press the button.


johnny-keker (approved):
LGTM

PR #167, opened 2020-09-29
Status file

johnny-keker:
Addresses #140
Addresses #141 

This PR expands the function of `status file`. Status file was used to provide some text that appears in `Debug Visualizer` after every debug run. Now it contains some dispatch parameters following this pattern:

```
grid size (512, 1, 1)
group size (64, 1, 1)
wave size 32
comment status string
```

In the example above following parameters are set:

* `grid size X` = 512, `grid size Y` = 1 and `grid size Z` = 1
* `group size X` = 64, `group size Y` = 1 and `group size Z` = 1
* `wave size` = 32
* `comment` = "status string"

This parameters affecting some project properties:

* `NDRange 3D` is set true if any of `grid size Y` and `grid size Z` is not zero
* `Group Size` is set according to `group size X` (3D group sizes is not supported yet)
* `DimX`, `DimY` and `DimZ` is set according to `grid size` / `group size` for perspective axis
* `NGroups` is equals `DimX` if `NDRange 3D` is false or as multiplication of all `Dim`'s otherwise
* `WaveSize` is set from status file which affects the magick number checker
  
In cases below you'll get corresponding error:

* The status file contents do not follow the right pattern
* Either `grid size X`, `group size X` or `wave size` equals zero
* `grid size Y` and `grid size Z` is not zero but `group size Y` or `group size Z` is
* Any `group size` is bigger than correspond `grid size`
* `wave size` is bigger than `group size X`

### DIFF
--- a/VSRAD.Package/Server/BreakState.cs
+++ b/VSRAD.Package/Server/BreakState.cs
@@ -37,11 +37,11 @@ namespace VSRAD.Package.Server
                 var groupY = uint.Parse(match.Groups["gp_y"].Value);
                 var groupZ = uint.Parse(match.Groups["gp_z"].Value);
 
+                NDRange3D = gridY != 0 && gridZ != 0;
                 GroupSize = groupX;
                 DimX = gridX / groupX;
-                DimY = gridY / groupY;
-                DimZ = gridZ / groupZ;
-                NDRange3D = gridY != 0 && gridZ != 0;
+                DimY = NDRange3D ? gridY / groupY : 0;
+                DimZ = NDRange3D ? gridZ / groupZ : 0;
                 WaveSize = uint.Parse(match.Groups["wv"].Value);
                 StatusString = match.Groups["comment"].Value;
             }
--- a/VSRAD.Package/Server/BreakState.cs
+++ b/VSRAD.Package/Server/BreakState.cs
@@ -15,6 +15,7 @@ namespace VSRAD.Package.Server
         public uint DimX { get; }
         public uint DimY { get; }
         public uint DimZ { get; }
+        public uint GroupSize { get; } // TODO: handle 3d group sizes
         public bool NDRange3D { get; }
 
         private static readonly Regex StatusFileRegex = new Regex(@"grid size \((?<gd_x>\d+), (?<gd_y>\d+), (?<gd_z>\d+)\)\s+group size \((?<gp_x>\d+), (?<gp_y>\d+), (?<gp_z>\d+)\)\s+wave size (?<wv>\d+)\s+comment (?<comment>.+)", RegexOptions.Compiled);
@@ -36,6 +37,7 @@ namespace VSRAD.Package.Server
                 var groupY = uint.Parse(match.Groups["gp_y"].Value);
                 var groupZ = uint.Parse(match.Groups["gp_z"].Value);
 
+                GroupSize = groupX;
                 DimX = gridX / groupX;
                 DimY = gridY / groupY;
                 DimZ = gridZ / groupZ;
--- a/VSRAD.Package/Server/BreakState.cs
+++ b/VSRAD.Package/Server/BreakState.cs
@@ -1,95 +1,21 @@
 ﻿using System;
-using System.Text.RegularExpressions;
-using VSRAD.Package.Utils;
 
 namespace VSRAD.Package.Server
 {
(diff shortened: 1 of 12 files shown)


VSRAD.Package/Server/BreakState.cs
```diff
@@ -10,14 +11,62 @@ public sealed class BreakState
...
+                var gridZ = uint.Parse(match.Groups["gd_z"].Value);
+                var groupX = uint.Parse(match.Groups["gp_x"].Value);
+                var groupY = uint.Parse(match.Groups["gp_y"].Value);
+                var groupZ = uint.Parse(match.Groups["gp_z"].Value);
+                WaveSize = uint.Parse(match.Groups["wv"].Value);
+                NDRange3D = gridY != 0 && gridZ != 0;
+
+                if ((NDRange3D && (groupY == 0 || groupZ == 0 || groupY > gridY || groupZ > gridZ))
+                    || groupX > gridX || WaveSize > groupX || groupX == 0 || WaveSize == 0 || gridX == 0)
+                    throw new ArgumentException();
```
Slimakanzer, line 45: More informative error messages should be given to the user. This validation can be divided into several parts, each of which gives the user a useful error message.

VSRAD.Package/Server/BreakState.cs
```diff
@@ -10,14 +11,62 @@ public sealed class BreakState
...
+
+                GroupSize = groupX;
+                DimX = gridX / groupX;
+                DimY = NDRange3D ? gridY / groupY : 0;
+                DimZ = NDRange3D ? gridZ / groupZ : 0;
```
Slimakanzer, line 50: Is this the correct division? What happens if the gridSize is not completely divisible by groupSize?

Slimakanzer (changes requested):
This picture shows that there is no space after _**status:**_

misc

VSRAD.Package/Server/BreakState.cs
```diff
@@ -10,14 +11,62 @@ public sealed class BreakState
...
+                StatusString = match.Groups["comment"].Value;
+            }
+            else if (!string.IsNullOrEmpty(statusFileContents))
+            {
+                throw new ArgumentException();
+            }
+        }
+
+        public static BreakState GetBreakState(BreakStateData breakStateData, long totalElapsedMilliseconds, string statusFileContents)
```
timlathy, line 59: It's better to return `Result<BreakState>` and show an error message from the calling code (which handles UI). You can make the constructor accept all fields and do the validation here, so you won't even need to use exceptions, just `return new Error("...")`.

VSRAD.PackageTests/DebugVisualizer/ComputedColumnStylingTests.cs
```diff
@@ -93,5 +93,28 @@ public void MagicNumberCheckTest()
             for (int i = 192; i < 256; i++)
                 Assert.True((styling.ColumnState[i] & ColumnStates.Inactive) != 0);
         }
+
+        [Fact]
+        public void MagicNumberWithNonConstatnWaveSizeTest()
```
timlathy, line 98: `Constatn`

johnny-keker: Thanks! Fixed in 3b1dfb7

VSRAD.Package/Server/BreakState.cs
johnny-keker: Thanks for the suggestion, implemented in 7c0b3f5

VSRAD.Package/Server/BreakState.cs
johnny-keker: That's true, thanks. Fixed in 7c0b3f5

johnny-keker:
> 
> 
> This picture shows that there is no space after _**status:**_
> 
> misc

Thanks, fixed in f00e33d

VSRAD.Package/Server/BreakState.cs
johnny-keker: Since this is `uint`'s the floating part of the result will be discarded (similar to `floor` operation). This behavior is expected, that way we discard the partial groups if they exist

Slimakanzer (approved):
LGTM

timlathy (approved):

PR #168, opened 2020-10-13
Wavemap

johnny-keker:
Added a new feature: `Wavemap`. It is available above the visualizer table, can be hidden just as `Columns`, `App Args` and `Break Args`.

`Wavemap` is a canvas that displays rectangles of different colors. Each rectangle corresponds to one wave. The color indicates the break line of its wave, so you can visually see which waves stepped on which breakpoints. When you hover your mouse over the rectangle the following info will appear in popup:

```
Group: X
Wave: X
Line: X
```

That is the info about what wave of what group is currently under the mouse and which breakline it has.

The breakline info is stored in the first lane of system watch (after a magic num) in every wave.

The wavemap is always shaped as `numGroups x wavesPerGroup` for convenience, so every column corresponds to one group executed.

You can change the size of rectangles in `Visualizer Appearance` tab.

### DIFF
--- a/VSRAD.Package/DebugVisualizer/Wavemap/WavemapCanvas.cs
+++ b/VSRAD.Package/DebugVisualizer/Wavemap/WavemapCanvas.cs
@@ -41,8 +41,8 @@ namespace VSRAD.Package.DebugVisualizer
                 for (int j = 0; j < 2; ++j)
                 {
                     var r = GetWaveRectangleByCoordinates(j, i);
-                    _canvas.Children.OfType<Rectangle>().ElementAt(j * 200 + i).Fill = r.Fill;
-                    _canvas.Children.OfType<Rectangle>().ElementAt(j * 200 + i).ToolTip = r.ToolTip;
+                    _rectangles[j][i].Fill = r.Fill;
+                    _rectangles[j][i].ToolTip = r.ToolTip;
                 }
             }
             _canvas.InvalidateVisual();
--- a/VSRAD.Package/DebugVisualizer/Wavemap/WavemapCanvas.cs
+++ b/VSRAD.Package/DebugVisualizer/Wavemap/WavemapCanvas.cs
@@ -1,15 +1,7 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Windows.Shapes;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+﻿using System.Windows.Shapes;
 using System.Windows.Controls;
 using System.Windows.Media;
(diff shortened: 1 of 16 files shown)


VSRAD.Package/DebugVisualizer/Wavemap/WavemapView.cs
```diff
@@ -0,0 +1,93 @@
...
+namespace VSRAD.Package.DebugVisualizer.Wavemap
+{
+    public struct WaveInfo
+    {
+        public Brush BreakColor;
+        public uint BreakLine;
+        public int GroupIdx;
+        public int WaveIdx;
+    }
+
```
Slimakanzer, line 14: It would be better to use `#pragma warning disable CA1815 ` here to turn off warning

Slimakanzer (changes requested):

VSRAD.Package/DebugVisualizer/Wavemap/WavemapView.cs
johnny-keker: Thanks, resolved in b3bd8d0

VSRAD.Package/DebugVisualizer/Wavemap/WavemapCanvas.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        private void UpdateWaveRectangle(int row, int column)
+        {
+            var r = _rectangles[row][column];
+            var validWave = _view != null && _view.IsValidWave(row, column);
+            r.ToolTip = new ToolTip()
+            {
+                Content = validWave ?
+                    $"Group: {_view[row, column].GroupIdx}\nWave: {_view[row, column].WaveIdx}\nLine: {_view[row, column].BreakLine}"
```
timlathy, line 129: It's better to assign `_view[row, column]` to a local variable instead of repeatedly calling the indexer, since it isn't a simple memory lookup, you have additional logic there.

johnny-keker: Thanks for suggestion, resolved in 2712ed9

Resolving this needed additional refactoring, since `[row, column]` getter did not check any boundaries, so if I just rewrote it as is, we could get out of bounds exception while trying to get break line of hidden group.

So, I generally improved `WavemapView` interface by making `IsValidWave` private, check it in getter and return default `WaveInfo` struct if requested wave is out of bounds of given data.

Slimakanzer (approved):
LGTM, works as expected

PR #163, opened 2020-09-10
Analysis refactoring

Slimakanzer:
* The logic of the document analysis has been completely changed
* Added a document manager that monitors the states of the RadAsm documents and contains them in the document table
* Added `InvisibleDocument` which loads in the background (example: included files, instruction list files and etc.)
* The search for navigation tokens has been changed, now it is completely based on the parser result

PR #169, opened 2020-10-14
gfx instruction set switch

Slimakanzer:
This PR adds a new instruction set selection feature.
The user can select the set of instructions (gfx) he needs. After selecting a set the autocomplete, highlighting and intellisence will work only with selected gfx.

Capture

This drop-down list can be found in two places:
* `View -> Toolbars -> Instruction sets`
* `View -> Toolbars -> RAD Debug` only if Radeon Asm Debugger is used

WIP:
- [x] Error highlight instructions not from the current set

Known issues:
- [x] Syntax highlighting does not update after set selecting


johnny-keker (changes requested):
The click on instruction in the context menu leads to a `NullReferenceException` in the `InvisibleDocument.cs:19`, that lead to crashing of VS instance

image

Slimakanzer:
@johnny-keker thank you for the error you found. This bug is more related to #163 (Analysis refactoring) PR.

Slimakanzer:
@johnny-keker thanks for your help, fixed f4dbffd

PR #170, opened 2020-10-20
Server reconnect attempt

johnny-keker:
Addresses #159 

If current connection is down, the extension will take one attempt to reconnect to server. That way you would not get an error if you rebooted server on remote machine.

### DIFF
--- a/VSRAD.Package/Server/CommunicationChannel.cs
+++ b/VSRAD.Package/Server/CommunicationChannel.cs
@@ -23,7 +23,7 @@ namespace VSRAD.Package.Server
 
         ClientState ConnectionState { get; }
 
-        Task SendAsync(ICommand command, bool tryReconnect = true);
+        Task SendAsync(ICommand command);
 
         Task<T> SendWithReplyAsync<T>(ICommand command) where T : IResponse;
 
@@ -85,7 +85,9 @@ namespace VSRAD.Package.Server
                 options.PropertyChanged += (s, e) => { if (e.PropertyName == nameof(options.ActiveProfile)) ForceDisconnect(); };
         }
 
-        public async Task SendAsync(ICommand command, bool tryReconnect = true)
+        public async Task SendAsync(ICommand command) => await SendAsync(command, true);
+
+        private async Task SendAsync(ICommand command, bool tryReconnect)
         {
             try
             {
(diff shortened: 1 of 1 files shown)


VSRAD.Package/Server/CommunicationChannel.cs
```diff
@@ -85,9 +85,21 @@ public CommunicationChannel(SVsServiceProvider provider, IProject project)
                 options.PropertyChanged += (s, e) => { if (e.PropertyName == nameof(options.ActiveProfile)) ForceDisconnect(); };
         }
 
-        public async Task SendAsync(ICommand command)
+        public async Task SendAsync(ICommand command) => await SendAsync(command, true);
```
timlathy, line 88: `async` is not required here, you can just return the task:
`public Task SendAsync(ICommand command) => SendAsync(command, true);`

VSRAD.Package/Server/CommunicationChannel.cs
```diff
@@ -85,9 +85,21 @@ public CommunicationChannel(SVsServiceProvider provider, IProject project)
...
+
+        private async Task SendAsync(ICommand command, bool tryReconnect)
         {
-            await EstablishServerConnectionAsync().ConfigureAwait(false);
+            try
```
timlathy, line 92: Handling `ConnectionRefusedException` here is not needed, since reconnecting won't help if the destination host is unreachable.

johnny-keker: This handler is needed to show user "Could not reconnect to server" message in server output window and showing "Connection to {ConnectionOptions} has been terminated:" error instead of "Unable to establish connection..."

Without it it's not that clear what happened from user's point of view. Note that because of `when (!tryReconnect)` check if server is down on a stage of establishing connection, the behavior would be same as before.

Maybe we should consider removing `else` on line 119 instead, since if server is actually down, we'll show error by handling `ConnectionRefusedException` and we'll actually never fall in this case?

VSRAD.Package/Server/CommunicationChannel.cs
johnny-keker: Thanks, fixed in 49b5022

VSRAD.Package/Server/CommunicationChannel.cs
timlathy: Okay, let's keep it.

Regarding line 119: the `else` branch is necessary because exceptions there can be unrelated to connection state. The call to `ForceDisconnect()` should be moved out of `else` and placed before the conditional, though. Consider serialization errors: we cannot use the same connection because we may have read the server response only partially, so calling `SendAsync` again without dropping the connection can lead to unexpected errors.

johnny-keker: Thanks for the suggestion, fixed in 40d8520

timlathy (approved):

PR #171, opened 2020-10-29
Evaluate $(RadBreakLine) in non-debug contexts

timlathy:
`$(RadBreakLine)`, which stands for the next breakpoint line in the current file, can now be used in all actions, not just `Debug`, and is correctly evaluated in the macro editor.

The underlying change that enables this is moving breakpoint tracking from the debug engine (`VSRAD.Deborgar`) to the main extension assembly (`VSRAD.Package`). Instead of keeping a collection of breakpoints added through [`IDebugEngine2.CreatePendingBreakpoint`](https://docs.microsoft.com/en-us/visualstudio/extensibility/debugger/reference/idebugengine2-creatependingbreakpoint), we use the [`EnvDTE`](https://docs.microsoft.com/en-us/dotnet/api/envdte.debugger.breakpoints) API.

This simplifies the interaction between the DE and the extension, which, in turn, makes it easy to fix an old known issue. Previously, starting the debugger when no files are open in the editor would show a cryptic error message ("unable to attach") and prevent you from debugging until you restart VS. Now it just shows the usual "No files open in the editor" warning.


johnny-keker (approved):
Done some testing:

* Used RadBreakLine macro in custom action
* Tried run debug with no files opened in editor
* Run custom action with RadBreakLine after debug run

LGTM, works as expected

PR #175, opened 2020-11-21
Replace hardcoded debug action with Read Debug Data step

timlathy:
* `DebuggerProfileOptions` are replaced with a step that performs the same function (reading valid watches/status/output data) and can be added to any action.
* Configuration files created in previous extension versions are automatically converted to the new format (a regular action with a Read Debug Data step is created based on legacy debugger options).
* An action with a Read Debug Data step can be run without starting the VS debugger or alongside a debugger for another language/platform.
* A custom _current statement_ arrow is shown in the editor margin next to the last break line (or multiple lines, if _Multiple active breakpoints_ mode is enabled).
* Debugger output can be read locally — running a debug server on `localhost` is no longer required.


johnny-keker (changes requested):
In a text mode with non-zero offset local debug data fetch fails with exception.

timlathy:
> In a text mode with non-zero offset local debug data fetch fails with exception.

Thanks for catching this! Fixed in 4e19a8c

johnny-keker (approved):

PR #176, opened 2020-12-10
Localhost switch in profile settings

### LINKED ISSUE #173: Localhost switch in profile settings
- Rename `Remote Port`->`DebugServer Port`
- Add `Localhost` switch in `Profiles`->`General`. When set:
   - `Remote Machine Address:` is greyd out
   - Value `localhost` is used for connetion

timlathy:
This PR:
* Resolves #173
* Adds a toolbar dropdown for changing remote machine address without going to the profile editor

### DIFF
--- a/VSRAD.Package/Commands/ProfileDropdownCommand.cs
+++ b/VSRAD.Package/Commands/ProfileDropdownCommand.cs
@@ -1,0 +1,77 @@
+﻿using Microsoft.VisualStudio.OLE.Interop;
+using Microsoft.VisualStudio.ProjectSystem;
+using System;
+using System.ComponentModel.Composition;
+using System.Linq;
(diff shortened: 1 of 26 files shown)


johnny-keker (changes requested):
Done some testing, got a little unexpected behavior:

1. Set up default profile host (for example 127.0.0.1)
2. Add new profile with different host
3. We can see current profile host in the dropdown, but alternatives are "Local" and "127.0.0.1"
4. Clicking on "127.0.0.1" actually overrides new profile host

timlathy:
Thanks, fixed in 375299b. Changing the active profile now forces the dropdown to be updated.

johnny-keker (approved):
Done some testing
LGTM, works as expected

Issue #138, opened 2020-08-17
Error lines mapping is broken


Kirpich30000:
Redesigned profiles broke error messages line mapping.

timlathy:
This issue is no longer applicable: line mapping was removed in 2d9703c because we decided it works better as a separate script, not part of the extension code.

Issue #21, opened 2019-11-30
Error handling issues


Kirpich30000:
- [x]  LineMapping is wrong. Presumably `SourceLine = int.Parse(match.Groups["line"].Value) - 1` should fix it
- [ ] Error messages might not have line numbers. Currently plugin doesn't process such situation correctly

timlathy:
This issue is no longer applicable: line mapping was removed in 2d9703ce45c998bc9d28e6d3817d1f42290bcfd3 because we decided it works better as a separate script, not part of the extension code.

PR #178, opened 2020-12-19
Minor UI fixes and code cleanup

timlathy:
* Renamed `Status File` to `Dispatch Parameters File`
* Fixed a crash when attempting to save profiles with conflicting names and pressing "Cancel" in the name conflict resolution dialog
* Removed preprocessor line mapping. This code was broken for a long time (see #138, #21) and can be reimplemented as a separate script
* Fixed current statement markers not showing up in Release builds


johnny-keker (approved):
LGTM, works as expected

PR #179, opened 2020-12-31
Profile editor changes

timlathy:
* Remove _Copy Sources to Remote_ (deploy)
* Rename _Local/Remote Working Directory_ to _Default Local/Remote Working Dir_
* Refer to _Default Working Dir_ in tooltips for _Working Directory_ in _Execute_ step and _Source/Target Path_ in _Copy File_ step
* When _Run on Localhost_ is enabled, show the original settings for _Copy File_, _Execute_, _Read Debug Data_ locations and a tooltip explaining that they're overridden by _Run on Localhost_


johnny-keker (changes requested):
Looks like the results of reading of the same binary data on local and remote are different
screen1
screen2

In this example I used the binary mode with offset 4

timlathy:
Thanks! It seems like there were two separate issues, both present only when binary output is on and output offset is > 0.

johnny-keker (approved):
Done some testing, LGTM, works as expected.

PR #182, opened 2021-01-26
Syntax fixes

Slimakanzer:
- [x] Fix incorrect outlining with if-elsif-endif preprocess blocks
- [x] Check incorrect Function List behavior
- [x] Check Function List on reloading VS
- [x] Highlight `return` asm2 keyword


johnny-keker (approved):
Done some testing:
* Opened multiple documents
* Closed and Opened VS with multiple documents
* Edited function names
* Navigated through document

LGTM, works as expected

PR #183, opened 2021-01-26
Add target hosts editor

timlathy:
* Clicking _Edit..._ at the bottom of the target hosts dropdown opens the editor window, where hosts can be added, edited or removed.
* Changes can be discarded by pressing _Cancel_ or closing the editor window and choosing _No_ in the _Save changes?_ dialog.
* The maximum number of hosts in the dropdown is no longer limited.

### DIFF
--- a/VSRAD.Package/Commands/ProfileDropdownCommand.cs
+++ b/VSRAD.Package/Commands/ProfileDropdownCommand.cs
@@ -7,8 +7,10 @@ using System;
 using System.ComponentModel.Composition;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Windows;
 using VSRAD.Package.Options;
 using VSRAD.Package.ProjectSystem;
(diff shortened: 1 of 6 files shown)


johnny-keker (approved):
Done some testing, LGTM, works as expected

VSRAD.Package/Commands/ProfileDropdownCommand.cs
```diff
@@ -67,7 +75,7 @@ private void ListTargetMachines(IntPtr variantOut)
             // (if the profile is not changed this is a no-op because the current host is already at the top of the list)
             _project.Options.RecentlyUsedHosts.Add(_project.Options.Profile.General.Connection.ToString());
```
Kirpich30000, line 76: It isn't RecentlyUsedHosts, not anymore. Suggest to rename it.

PR #189, opened 2021-03-28
Support transferring directories using Copy File

### LINKED ISSUE #114: Replace hardcoded Copy Sources logic with Copy File step
- [ ] Support transferring directories using Copy File (rename the step to "Copy File/Directory")
  * It's better to reuse an existing step to avoid duplicating UI and validation logic
- [ ] Add a "skip if not modified" option to Copy File
  * When sending a file for the first time, record its last modification date, and don't resend it unless the timestamp changes
  * For directories, keep a timestamp for each file, so we don't resend the whole directory when a single file is changed
  * Do we need to add it for Remote -> Local transfers too?

"Check source timestamp" can be replaced with a "Check modification time" dropdown with the following options:
* "Don't check" (current behavior when "check source timestamp" is off)
* "Fail if not modified" (current behavior when "check source timestamp" is on)
* "Skip if not modified" (new behavior)

timlathy:
This PR resolves #114:
* *Copy File* step can now copy directories.
* Files with identical sizes and modification times can be skipped (*Skip if not modified*).
* Compression can be enabled for Remote->Local and Local->Remote copies (*Use compression*). This setting is ignored for Local->Local transfers.
* Modification times of destination files can be set to modification times of source files (*Preserve timestamps*).


timlathy:
Manual testing will be performed as a part of the review of the (upcoming) debug server capabilities PR, since the latter includes a partial rewrite of the file copy logic enabled by server capability info.

PR #190, opened 2021-04-03
Debug server capabilities check

### LINKED ISSUE #160: Ask Debug Server for supported capabilities, report if a newer server version is required
Add a  `GetCapabilities` command to request capabilities (command IDs) supported by the server. When changing the serialization format of an existing command (adding/removing fields), we'll add another command with a new ID to preserve compatibility with previous extension versions.

If the server does not recognize `GetCapabilities`, we assume it's an old version and reconnect without sending the command if the old capabilities are sufficient.

This way:
* the server keeps supporting old extension versions (because the format of existing commands does not change)
* new extension versions can have new commands added while showing a helpful message if the server does not support them (currently the connection is just reset)
* new extension versions can work in a reduced functionality mode if the server does not support certain commands (?)

timlathy:
This PR closes #160. See the issue for an overview of the capability checking mechanism.

In addition to the server capabilities check, this PR includes a change to how paths in action steps are resolved. Previously, remote paths that were relative to the working directory (`$(RemoteWorkDir)`) could not be converted to absolute paths on the extension side because we couldn't tell whether the remote host uses Windows or Unix paths. Now the server identifies its platform along with capabilities, so we can join paths on the VS side and send just the absolute path in IPC commands.

### DIFF
--- a/VSRAD.DebugServer/Dispatcher.cs
+++ b/VSRAD.DebugServer/Dispatcher.cs
@@ -17,6 +17,7 @@ namespace VSRAD.DebugServer
             PutDirectoryCommand pd => new PutDirectoryHandler(pd).RunAsync(),
             ListFilesCommand lf => new ListFilesHandler(lf).RunAsync(),
             GetFilesCommand gf => new GetFilesHandler(gf).RunAsync(),
+            GetServerCapabilitiesCommand _ => new GetServerCapabilitiesHandler().RunAsync(),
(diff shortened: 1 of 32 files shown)


johnny-keker (approved):
Done some testing:

* Doesn't work with the old Debug Server, shows appropriate error;
* Works with new Debug Server;
* The paths is converted to a absolute paths on the VS side.

LGTM

PR #191, opened 2021-04-12
Obsolete config format import

johnny-keker:
Renamed `.conf.json` to `.profiles.json` so old config would not be overwritten.

This PR adds the ability to successfully run the project with old `.conf.json` config. After first updating of configs, the configs will be updated to match the new `.profiles/.user` format, that makes the migration to new configs format seamless for user.


timlathy (approved):
I tested this on several legacy configuration files without experiencing any problems. Commit 146e005 has resolved all my concerns wrt code readability and a regression that could cause NRE (`Note that DeserializeObject can return null even on success`), looks good to me now.

PR #193, opened 2021-04-19
Syntax signature help

Slimakanzer:
## Description

This PR provides signature help pop-up window with function/instruction signature.
Also highlighted current parameter when moving the caret or changed parameters.

Signature help triggered to the right of the function/instruction reference:
- *Asm1*: `space` symbol  for both
- *Asm2*: `(` symbol for the functions and `space` symbol for the instructions

⚠ To enable feature go to `Tools -> Options -> RadeonAsm Options -> Intellisense, Signature help` and select *True*

## Example
image


johnny-keker (commented):
I believe it would be more convenient for user if signature help popup would close immediately after changing function name. Now I have to insert newline/press Esc for it to disappear or change function name to another function for it to reflect the new signature.

Slimakanzer:
@johnny-keker thank you, fixed in e2eaaf7

johnny-keker (approved):
Done some testing, after e2eaaf7 and 98901ec things look great!

LGTM

PR #194, opened 2021-04-23
Improve handling of long-running/hanged actions

timlathy:
This PR allows the currently running action to be cancelled at any point in time. This is mostly useful when an `Execute` step starts a process that hangs or takes too long to finish. Previously, this behavior was available only for remote actions and had negative side-effects: _cancelling_ meant force-disconnecting from the server. This effectively terminated the action on the VS side but didn't stop the server from still running the command (e.g. waiting for the process to exit), which, coupled with the fact that commands from different clients cannot execute simultaneously, could cause the server to become unresponsive. To prevent this situation, a bidirectional ping mechanism is added. Currently it's only used by the server in the `Execute` command handler to detect when a client disconnects and terminate the process.

To preserve backward compatibility, the capability exchange (when the extension connects to the server) now includes extension capabilities as well. This way, the server knows if the client supports new behaviors (like pings) and can use fallbacks otherwise.

For local actions, cancellation is currently only supported by the process observer code. When a local `Execute` step hangs, cancelling the action will terminate the process. In future, cancellation can support can be added to other local steps as well, though I'm not aware of situations where they can take too long to complete.

In addition to that, this changes the behavior of the `Execute` step to show a confirmation dialog before terminating the process (with an option to wait for the process to exit). This requires new commands/responses:
* `ExecutionTimedOutResponse`, sent by the server when a timeout is reached, contains a list of processes that timed out (shown in the confirmation dialog);
* `ExecutionTimedOutActionCommand`, sent by the extension to indicate what action needs to be taken (terminate the process/wait);
* `ExecutionTerminatedResponse`, sent by the server in response to `ExecutionTimedOutActionCommand { TerminateProcesses = true }`, contains a list of processes that were terminated.

The server does not use these responses if the extension capabilities do not include `Base`, indicating a legacy extension version. In this case, the default behavior of terminating processes immediately is assumed, and instead of `ExecutionTerminatedResponse` the server sends `ExecutionCompleted { Status = TimedOut }`.

The old behavior can be enabled in the `Execute` step settings by setting `Timeout Behavior` to `Terminate`.


johnny-keker (approved):
Done some testing. Action canceling seems to work fine, so as subprocesses handling.

LGTM, works as expected

Issue #144, opened 2020-08-21
Function arguments helper


Kirpich30000:
We need to enable following VS feature:
![image](https://user-images.githubusercontent.com/12259057/90862611-5c381400-e396-11ea-85aa-759777700147.png)

Kirpich30000:
instruction helpers (+switch in options) would be also nice to have

PR #195, opened 2021-04-26
IntelliSense syntax fixes

Slimakanzer:
- [x] Fix exception on _Peek Definition_
- [x] Return the definition description feature (it was temporarily disabled)
- [x] Dismiss signature help popup on word deletion (ctrl+backspace), line deletion (ctrl+shift+l), delete to Begin Of Line, delete to End Of Line (improve #144 usability)


johnny-keker (approved):
Done some testing:

* Peek definition works as expected;
* Definition description showed;
* Signature helper dismissed after using shortcuts.

LGTM

PR #198, opened 2021-04-26
Visualizer improvements

johnny-keker:
# Added:

* New feature `Append leading zeroes to binary/hex numbers` available in Visualizer Appearance tab inside Options window.
When active, it will make hex numbers presented as 8 digits (`0x00000001`) and binary as 32 digits.
* New feature `Insert separator in binary/hex mode between groups of X digits` available in Visualizer Appearance tab inside Options window. When `X` is not zero, it will insert whitespaces each `X` characters in bin/hex representations starting from last digit (`0x4 23 13`).
* New feature `Insert separator in int/uint mode between groups of X digits` available in Visualizer Appearance tab inside Options window. When `X` is not zero, it will insert whitespaces each `X` characters in int/uint representations starting from last digit (`1 000 000`).
* Custom array slices for `Add To Watches As Array` feature. In context menu after all predefined starting indexes appeared new button with label `Custom`. Hitting it will open dialog window with `Start`, `Step` and `Count` input fields and `OK`/`Cancel` buttons. After clicking `OK`, the slice defined by input fields will be added to watch list.

# Changed:

* `Vertical separator between each X columns` setting moved to the Visualizer Appearance tab.

### DIFF
--- a/VSRAD.Package/Commands/AddToWatchesCommand.cs
+++ b/VSRAD.Package/Commands/AddToWatchesCommand.cs
@@ -34,6 +34,13 @@ namespace VSRAD.Package.Commands
             return OLECMDF.OLECMDF_ENABLED | OLECMDF.OLECMDF_SUPPORTED;
         }
 
+        private void HandleCustomSlice(uint start, uint step, uint count, string watchName)
+        {
+            var arrayRangeWatch = ArrayRange.FormatCustomSlice(watchName, (int)start, (int)step, (int)count);
+            foreach (var watch in arrayRangeWatch)
+                _toolIntegration.AddWatchFromEditor(watch);
+        }
+
         public void Execute(uint commandId, uint commandExecOpt, IntPtr variantIn, IntPtr variantOut)
         {
             var watchName = _codeEditor.GetActiveWord();
@@ -46,7 +53,7 @@ namespace VSRAD.Package.Commands
             }
             else if (commandId == Constants.AddToWatchesArrayCustomCommandId)
             {
-
+                new AddToWatchesCustomSliceEditor(HandleCustomSlice, watchName) { ShowInTaskbar = false }.ShowModal();
             }
             else if (commandId >= Constants.AddArrayToWatchesToIdOffset)
             {
--- a/VSRAD.Package/Commands/AddToWatchesCommand.cs
+++ b/VSRAD.Package/Commands/AddToWatchesCommand.cs
@@ -44,6 +44,10 @@ namespace VSRAD.Package.Commands
             {
                 _toolIntegration.AddWatchFromEditor(watchName);
             }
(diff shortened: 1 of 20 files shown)


timlathy (commented):
All new functionality appears to work correctly, but I have a couple of (small) suggestions:

First, I'd rename `Insert separator in * mode between each * numbers` to `Insert separator in * mode between groups of * digits`.

Next, `AddToWatchesCustomSliceEditor` could be more keyboard-friendly. The input fields should come first in the tab order (try pressing Tab after opening the window -- it appears that OK/Cancel are before them). Then I'd put `Loaded += (s, e) => MoveFocus(new TraversalRequest(FocusNavigationDirection.First));` in the constructor to set focus on the first input when the window is opened. Also if you add `IsDefault="True"` to the OK button definition in XAML, the dialog can be accepted by hitting Enter after filling in the inputs.

Finally, I think it'd be better if `NumberInput` selected all text when it receives keyboard focus. To achieve this, the `TextBox` definition in `NumberInput.xaml` needs `GotKeyboardFocus="...event handler..."`, where the event handler does something like:
```
void ...event handler...(object sender, ... e)
{
    var tb = (TextBox)sender;
    tb.Dispatcher.BeginInvoke(new Action(tb.SelectAll));
}
```

johnny-keker:
> 
> 
> All new functionality appears to work correctly, but I have a couple of (small) suggestions:
> 
> First, I'd rename `Insert separator in * mode between each * numbers` to `Insert separator in * mode between groups of * digits`.
> 
> Next, `AddToWatchesCustomSliceEditor` could be more keyboard-friendly. The input fields should come first in the tab order (try pressing Tab after opening the window -- it appears that OK/Cancel are before them). Then I'd put `Loaded += (s, e) => MoveFocus(new TraversalRequest(FocusNavigationDirection.First));` in the constructor to set focus on the first input when the window is opened. Also if you add `IsDefault="True"` to the OK button definition in XAML, the dialog can be accepted by hitting Enter after filling in the inputs.
> 
> Finally, I think it'd be better if `NumberInput` selected all text when it receives keyboard focus. To achieve this, the `TextBox` definition in `NumberInput.xaml` needs `GotKeyboardFocus="...event handler..."`, where the event handler does something like:
> 
> ```
> void ...event handler...(object sender, ... e)
> {
>     var tb = (TextBox)sender;
>     tb.Dispatcher.BeginInvoke(new Action(tb.SelectAll));
> }
> ```

Thanks for the suggestions! Implemented in 0f46db9

timlathy (approved):

Slimakanzer:
@johnny-keker 
 I apologize for the late reply, but I have a suggestion: `Step: 0` is meaningless it is better to set `1` immediately after opening.

image

PR #201, opened 2021-05-11
Ux improvements

johnny-keker:
Changed:

* Renamed `Break Args` field to `App Args 2`, `RadDebugBreakArgs` macro to `RadDebugAppArgs2`;
* Previously, any `Open in editor` step would change the active document window to newly opened one even if this document is opened in a separate tab. Now the active document will remain the same as it was before running of the action. If `Line Marker` present, the corresponding line will be active in opened file once user change the tab to it.
* Previously, `Add to watches` feature selected active word including indices (`var[a][b][c]`), now it ignores them (in previous example just `var` will be added to the watch list).

Added:

* `App Args 3` field and corresponding macro `RadDebugAppArgs3`.

Fixed:

* NRE if old config file didn't had `TargetHosts` field

### DIFF
--- a/VSRAD.Package/Utils/VsEditor.cs
+++ b/VSRAD.Package/Utils/VsEditor.cs
@@ -53,7 +53,9 @@ namespace VSRAD.Package.Utils
             {
                 var dte = serviceProvider.GetService(typeof(DTE)) as DTE;
                 Assumes.Present(dte);
+                var activeFile = dte.ActiveDocument;
                 dte.ItemOperations.OpenFile(path);
+                dte.ItemOperations.OpenFile(activeFile.FullName); // preserving old active document
 
                 if (string.IsNullOrEmpty(lineMarker))
                     return;
@@ -63,6 +65,7 @@ namespace VSRAD.Package.Utils
                 var textManager = serviceProvider.GetService(typeof(SVsTextManager)) as IVsTextManager2;
                 Assumes.Present(textManager);
 
+                dte.ItemOperations.OpenFile(path); // change active document if line marker set
                 textManager.GetActiveView2(1, null, (uint)_VIEWFRAMETYPE.vftCodeWindow, out var activeView);
                 activeView.SetCaretPos(lineNumber, 0);
             }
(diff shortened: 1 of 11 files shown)


Kirpich30000:
That's bad if the active file depends on LineMarker. It's better to always preserve current active file. LineMarker (if present) should only scroll, but not affect which document stays active at the end.

Kirpich30000 (changes requested):
See comment above

johnny-keker:
> 
> 
> That's bad if the active file depends on LineMarker. It's better to always preserve current active file. LineMarker (if present) should only scroll, but not affect which document stays active at the end.

Thanks, fixed in 70c80a3

VSRAD.Package/Utils/VsEditor.cs
```diff
@@ -63,8 +65,10 @@ public static void OpenFileInEditor(SVsServiceProvider serviceProvider, string p
                 var textManager = serviceProvider.GetService(typeof(SVsTextManager)) as IVsTextManager2;
                 Assumes.Present(textManager);
 
+                dte.ItemOperations.OpenFile(path); // change active document to set active line according to line marker
```
Kirpich30000, line 68: Too many OpenFile(). Please restructure the code.

Kirpich30000 (changes requested):

VSRAD.Package/Utils/VsEditor.cs
johnny-keker: Done in f5158ad

Kirpich30000 (approved):
LGTM. Please do extensive testing of the changes before the merge.

Slimakanzer (approved):
LGTM, done some testing:
* App Args, App Args 2, App Args 3 fields
* `Open in editor` works as expected
* New `Add to watches` behavior

Found a little bug in `Open in editor` step. Visual studio tab with file content is not updated. @Kirpich30000  Is it critical?
To reproduce:
* Firstly run `Open in editor`  to open target file
* Edit file content, but **do not save**
* Run again `Open in editor`. The result: target file content is not changed in visual studio tab

Kirpich30000:
Is it updated if the changes are saved?

Slimakanzer:
@Kirpich30000 Yes, it works correctly in such a case

Kirpich30000:
Ok to merge

PR #202, opened 2021-05-13
Fix opened files error

Slimakanzer:
Fixed a error with saving opened files if some of them are presented in Visual Studio as tabs, but the files have already been removed from the file system.


Kirpich30000 (approved):
LGTM. If the proper testing is done then it's ok to merge.

timlathy (approved):

PR #206, opened 2021-05-17
Don't use DTE for opening files in VS editor

timlathy:
`IVsUIShellOpenDocument` provides a lower-level API, which seems to be faster. This mainly affects the `OpenInEditor` step.

### DIFF
--- a/VSRAD.Package/Utils/VsEditor.cs
+++ b/VSRAD.Package/Utils/VsEditor.cs
@@ -1,6 +1,7 @@
-﻿using EnvDTE;
-using Microsoft;
+﻿using Microsoft;
+using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TextManager.Interop;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


Slimakanzer:
Is there any other way to find line marker without reading the file a second time? [IVsTextLines](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.textmanager.interop.ivstextlines?view=visualstudiosdk-2019) structure returned by `IVsTextView.GetBuffer()` can't help with such a case?

https://github.com/vsrad/radeon-asm-tools/blob/4d853b7e02b5e345e61047b567fffa1d1c024600/VSRAD.Package/Utils/VsEditor.cs#L83

timlathy:
> 
> 
> Is there any other way to find line marker without reading the file a second time? [IVsTextLines](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.textmanager.interop.ivstextlines?view=visualstudiosdk-2019) structure returned by `IVsTextView.GetBuffer()` can't help with such case?
> 
> https://github.com/vsrad/radeon-asm-tools/blob/4d853b7e02b5e345e61047b567fffa1d1c024600/VSRAD.Package/Utils/VsEditor.cs#L83

Thanks for the suggestion, I implemented it in 53a03c2be03b56fca55edb76c43ade27ae0c5079. I didn't measure any significant performance difference even for large files; besides, this operation took on the order of <10ms to begin with. However, I think using `IVsTextLines` has one major advantage in that it prevents IO errors/possible data races: we search for the line marker in exactly the same lines that are shown in the editor.

PR #207, opened 2021-05-20
Fix syntax highlighting "disappearing" when the file is reloaded

timlathy:
As noted in the code, sometimes `ContentChanged` is raised when there are no changes to the buffer snapshot. This usually happens when the file is reloaded from disk but the content is unchanged (e.g. when using Disassemble/Preprocess actions from VSRAD.Package). The snapshot version in this case, however, gets incremented, so when `AnalysisClassifier.GetClassificationSpans` is called later, the early return condition (`analysisResult.Snapshot != span.Snapshot`) triggers, and the highlighting is no longer shown in the editor.


Slimakanzer:
LGTM!

PR #208, opened 2021-05-21
Add default keybindings for disassemble and profile actions

timlathy:
`Alt+D`: Disassemble, `Alt+W`: Profile

The keybindings can be changed in _Tools_ -> _Options_ -> _Keyboard_.


Kirpich30000 (approved):
LGTM

PR #210, opened 2021-06-01
MRU Interface Redesign

johnny-keker:
Moved Most Recently Used toggle button to the left side of control.

image


Kirpich30000 (approved):
lgtm

PR #214, opened 2021-06-07
Expand asm2 syntax keywords

Slimakanzer:
This PR adds `assigned`, `label` and `print` keywords to the asm2 syntax.
They are highlighted in the text with a default color for keywords.


Kirpich30000 (approved):
lgtm

PR #215, opened 2021-06-07
Fix intellisense autocomplete

Slimakanzer:
This PR fixes autocomplete trigger behavior. Now autocomplete will not trigger with **function** and **variable** declaration such as:
```
                                     
.macro someMacro
                ^ 
    not triggered at this position
.endm


.set someVariable
                 ^
    not triggered at this position
```


Kirpich30000 (approved):
lgtm

PR #217, opened 2021-06-08
Fix syntax highlighting

Slimakanzer:
This PR fixes syntax highlighting due to wrong token search boundaries. Especially this problem appeared if the token goes before the `if-elsif-endif` block such as:

```
.macro someMacro
.if someIf
.endif
.endm
```


Kirpich30000 (approved):
lgtm

PR #221, opened 2021-06-15
Function List: sorting and filtering

Slimakanzer:
This PR fixes `Function List` sorting and filtering in the case when both filters (_Text filter_ and _Type filter_) were applied.


johnny-keker (approved):
Done some testing.

Multiple filters bugs (for example, name filter overrides type filter) seems to be resolved.

Sorting works as expected.

LGTM

Slimakanzer:
via #252

PR #222, opened 2021-06-19
Update example

Slimakanzer:
This PR updates example vector-add application:
- Updated shader metadata to code-object-v3 version
- Updated VS project configs to match latest format
- Updated dockerfile to 2021.05.19
- Fixed some bugs in the VectorAdd HSA backend


johnny-keker:
LGTM, however I'd like to request some additional features.

* `README` file needs to be updated:
  * macro `RadDebugWorkDir` is now called `RadRemoteWorkDir`;
  * there is no `line 79` in the current version of `fp32_v_add.s`;
  * debugger invocation works different in the current version of extension.
* Since in the current version of extension user needs to manually bind Debug action to the corresponding button in the toolbar, it can be done in the project config.
* We can make use of the deploy feature and make a new action, that will deploy the `VectorAddProjectExample` to the remote machine and build it, so user won't do it manually.
* There is common problem that occurs when trying to debug from windows host: due to default CRLF line endings building of the shader fails. Fixing it requires manually changing the line endings in files `breakpoint_gcnasm.pl`, `dbg_clang_wrapper.sh` and `vadd_debug.sh` and replacing them on the remote machine (which can be done automatically using deploy feature). This scenario can be metioned in `README`.

Slimakanzer:
Suppressed by #233

PR #224, opened 2021-06-22
Add to Watches optional brackets matching

johnny-keker:
Added new `Match brackets on Add To Watches` property in Options window, that specifies whether brackets (`[]`) and their contents should be ignored on `Add to Watches` operation.


Kirpich30000 (approved):
LGTM

PR #225, opened 2021-06-22
Fix error on attempt to save deleted files

johnny-keker:
Fixed a error with saving opened files if some of them are presented in Visual Studio as tabs, but the files have already been removed from the file system.


Kirpich30000 (approved):
LGTM

PR #226, opened 2021-06-25
Fix memory leaks

Slimakanzer:
This PR fixes syntax extension memory leaks.
In fact, this PR contains three independent parts (one commit for each part), so the reviewer has luxury to skip what he is not interested in:
- Fix memory leak when closing document. 
Basically, this is an unsubscribe from document events when it is actually closed.
- Fix memory leak when closing document window. 
Basically, this is an unsubscribe from document window events when it is closed.
- Fix unreasonable memory usage by the lexer/parser.
Previously, for each document, an object of the parser class was created, which stored a copy of the instruction set. Now lexer/parser are singletons.

### DIFF
--- a/VSRAD.Syntax/Core/LexerParserFactory.cs
+++ b/VSRAD.Syntax/Core/LexerParserFactory.cs
@@ -1,4 +1,5 @@
-﻿using VSRAD.Syntax.Core.Lexer;
+﻿using System.Collections.Generic;
+using VSRAD.Syntax.Core.Lexer;
 using VSRAD.Syntax.Core.Parser;
 using VSRAD.Syntax.Helpers;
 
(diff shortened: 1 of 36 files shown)


VSRAD.Syntax/FunctionList/FunctionListProvider.cs
```diff
@@ -116,26 +130,30 @@ private void ActiveDocumentChanged(IDocument activeDocument)
         private void ClearFunctionList()
         {
             _functionListControl.ClearList();
```
timlathy, line 132: `_functionListControl` can be null at this point if the function list hasn't been open yet. When closing a document, we'll arrive here from `DocumentDisposed` and may hit a `NullReferenceException`. I suggest adding a null check.

VSRAD.Syntax/SyntaxHighlighter/ErrorHighlighter/SyntaxErrorHighlighterTaggerProvider.cs
```diff
@@ -25,7 +25,7 @@ public SyntaxErrorHighlighterTaggerProvider(IDocumentFactory documentFactory)
             var document = _documentFactory.GetOrCreateDocument(buffer);
             if (document == null) return null;
 
-            return new SyntaxErrorHighlighterTagger(document.DocumentAnalysis) as ITagger<T>;
+            return new SyntaxErrorHighlighterTagger(document) as ITagger<T>;
```
timlathy, line 28: For some reason, `SyntaxErrorHighlighterTaggerProvider.CreateTagger` and `ErrorHighlighterTaggerProvider.CreateTagger` are called multiple times with the same `ITextBuffer` instance. Each time a document is open, a breakpoint placed here is hit three times. If you take a heap snapshot after opening several documents, you can see that there are more instances of `SyntaxErrorHighlighterTagger` and `ErrorHighlighterTagger` than there are of other taggers.

Replacing this line with the following solves the problem:
```
return buffer.Properties.GetOrCreateSingletonProperty(() => new SyntaxErrorHighlighterTagger(document) as ITagger<T>);
```

I couldn't pinpoint the exact conditions when `CreateTagger` is called (and the documentation does not mention this), so I think it's possible that multiple instances per buffer can be created for other taggers as well, the problem is just easier to reproduce with these two. I suggest you to use `GetOrCreateSingletonProperty` here and in other `CreateTagger` implementations.

VSRAD.Syntax/Core/LexerParserFactory.cs
```diff
@@ -14,32 +15,50 @@ public struct LexerParser
...
     internal partial class DocumentFactory
     {
+        private Dictionary<AsmType, LexerParser> _asmSpecificLexerParser;
+
         private LexerParser? GetLexerParser(AsmType asmType)
         {
-            var instructionManager = _instructionManager.Value;
+            InitializeLexerParser();
+
+            if (_asmSpecificLexerParser.TryGetValue(asmType, out var lexerParser))
```
timlathy, line 24: (Subjective) I think that keeping the `switch` as it used to be and just replacing the `new` calls with `.Instance` accessors is a clearer approach that results in less code and less runtime overhead.

VSRAD.Syntax/FunctionList/FunctionListProvider.cs
Slimakanzer: Thank you, fixed in 80f866a

VSRAD.Syntax/SyntaxHighlighter/ErrorHighlighter/SyntaxErrorHighlighterTaggerProvider.cs
Slimakanzer: Thank you! I assumed that the VS under the hood handles such a situation (for example peek definition creates new View for the same Buffer) and manages resources correctly, but as it turns out, it doesn't. 

Fixed in b08ab7f

VSRAD.Syntax/Core/LexerParserFactory.cs
Slimakanzer: Thank you, I agree with you.
Fixed in 4dae221

PR #231, opened 2021-06-30
Invisible document

Slimakanzer:
This PR provides a better solution for invisible (included) documents.

Prior to this, the Text Buffer substitution approach was used. 
For example, `include.inc` was included using `#include "include.inc"`. A text buffer was created for the file, which was not controlled by Visual Studio in any way, but stored useful information for the parser. When opening this file (for example _GoTo Definition_) the text buffer was replaced with the text buffer provided by Visual Studio.

PR fixes this behavior. Now a single text buffer is created for the file, which is initially managed by the visual studio.

### What is the profit?
- Thanks to this, there is no need to parse the text a second time when opening the file (when replacing the buffer).
- Also a bug was fixed. If included file was opened (for example _GoTo Definition_) in the Visual Studio and then closed, then navigation to this file (_GoTo/Peek Definition_) did not work.


johnny-keker (approved):
LGTM

PR #234, opened 2021-07-09
Made Example C++ project

johnny-keker:
## Changed

* This PR changes the Example project type from `Radasm` to `C++` since we have the support for the latter and want to get rid of former in the future.

* Updated example `README` file to represent the current way of debugging.

## Fixed

* `Permission denied` error on debug of example app
* CRLF line endings error


Slimakanzer (approved):
LGTM

PR #237, opened 2021-07-13
Break Mode combobox for toolbar

johnny-keker:
This PR adds Break Mode selector combobox to the toolbar and removed same control from the Options window.

Screenshot 2021-07-15 173652


VSRAD.Package/Commands/BreakModeDropdownCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+                if (variantOut != IntPtr.Zero) /* list available items */
+                {
+                    Marshal.GetNativeVariantForObject(Utils.BreakModeConverter.BreakModeOptions, variantOut);
+                }
+            }
+            if (commandId == Constants.BreakModeDropdownId)
+            {
+                if (variantOut != IntPtr.Zero) /* get current item */
+                {
+                    var currentMode = Utils.BreakModeConverter.BreakModeToString(_project.Options.DebuggerOptions.BreakMode);
```
timlathy, line 44: Suggestion: add a separate array for "short" break mode strings (`{ "Round-robin", "Rerun line",  "Multiple" }`) and return them for `Constants.BreakModeDropdownId`. This way, you can set `defaultWidth` for the dropdown in .vsct to 90 instead of 240 without losing information: 
image

VSRAD.Package/ToolWindows/OptionsControl.xaml.cs
```diff
@@ -109,20 +109,7 @@ private void AlignmentButtonClick(object sender, RoutedEventArgs e)
 
     public sealed class BreakModeConverter : IValueConverter
```
timlathy, line 110: Suggestion: merge this into `Utils.BreakModeConverter`

VSRAD.Package/Utils/BreakModeConverter.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace VSRAD.Package.Utils
+{
+    static class BreakModeConverter
+    {
+        static public readonly string[] BreakModeOptions = new[] {  "Single active breakpoint, round-robin",
+                                                                    "Single active breakpoint, rerun same line",
+                                                                    "Multiple active breakpoints" };
+
+        public static String BreakModeToString(Options.BreakMode value)
```
timlathy, line 11: Now that we have an array of display names, we can use the enum value to index into it: `BreakModeOptions[(int)value]`.

timlathy (changes requested):

timlathy (approved):

PR #238, opened 2021-07-13
[WIP] Fix tokenization of preprocessor keyword

Slimakanzer:
This PR fixes preprocessor keyword highlighting.

## Issue
Lexer result:
```
#i
^ CHARP
  ^ IDENTIFIER
```

After the preprocessor keyword is fully written: 
```
#if
```
The tokenizer will apply changes (changed text: `if`) only to the _IDENTIFIER_ token, because text changes intersect only with this token, but not with CSHARP. 

After that, lexer result for the changed text will be:
```
#if
^ CHARP
  ^ IF keyword
```

## Solution (TBD)
One of the easiest and most understandable solutions is to apply the lexer not only to the changed text, but also to the whole line.

The second one is remove `#` token, but add `CLOSURE_IDENTIFIER` token (see [commit](https://github.com/vsrad/radeon-asm-tools/commit/3220981fc26a056cb028141a395cc43f70db37d4)).  In this case, it is not necessary to apply the lexer to the whole line.

### DIFF
--- a/VSRAD.Syntax/Core/DocumentTokenizer.cs
+++ b/VSRAD.Syntax/Core/DocumentTokenizer.cs
@@ -67,7 +67,7 @@ namespace VSRAD.Syntax.Core
                 // in some cases the text buffer may cause ContentChanged with 0 changes
                 if (args.Changes.Count == 0) return;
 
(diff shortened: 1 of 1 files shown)


Slimakanzer:
Via #239

PR #239, opened 2021-07-16
Fix tokenization of preprocessor keyword

Slimakanzer:
This PR fixes preprocessor keyword highlighting.

## Issue
Lexer result:
```
#i
^ CHARP
  ^ IDENTIFIER
```

After the preprocessor keyword is fully written (`#if`), tokenizer will apply changes (changed text: `if`) only to the _IDENTIFIER_ token, because text changes intersect only with this token, but not with CSHARP. 

After that, lexer result for the changed text will be:
```
#if
^ CHARP
  ^ IF keyword
```

## Solution
The solution is to remove `SHARP` (**#** symbol) token, but add `CLOSURE_IDENTIFIER` token. 
In this case lexer result will be:
```
#i
^ CLOSURE_IDENTIFIER
``` 

And after adding the letter `f`, tokenizer will apply the changes to the `CLOSURE_IDENTIFIER` token. And lexer result will be:
```
#if
^ PP_IF keyword
```


johnny-keker (approved):
Done some testing, LGTM

PR #240, opened 2021-07-16
Open navigation list on peek definition

Slimakanzer:
This pull request provides automatic opening of the navigation list at _Peek Definition_ action, if the number of navigation token is more than 1.

## Example
Some instructions have more than one declarations (`v_cndmask_b32`).
Now if you use Peak Definition on the `v_cndmask_b32` instruction, a navigation list with all declarations will be opened


johnny-keker (approved):
LGTM

PR #243, opened 2021-07-20
Options refactoring

Slimakanzer:
This PR doesn't fix anything, but it cleans up the options module.
The options have been split into two components:
- Option Page
- Option Model

The first is responsible for presenting options on the dialog page.
The second is responsible for saving and loading options from the Visual Studio registry.
This makes the options more expandable.

This PR may also slightly speed up the extension loading.

### DIFF
--- a/VSRAD.Syntax/FunctionList/FunctionListControl.xaml.cs
+++ b/VSRAD.Syntax/FunctionList/FunctionListControl.xaml.cs
@@ -1,5 +1,5 @@
 ﻿using VSRAD.Syntax.FunctionList.Commands;
-using static VSRAD.Syntax.Options.GeneralOptionPage;
+using VSRAD.Syntax.Options;
 using Microsoft.VisualStudio.Shell;
 using System;
 using System.Collections.Generic;
@@ -10,7 +10,6 @@ using System.Windows.Input;
 using Task = System.Threading.Tasks.Task;
 using System.Threading;
 using System.Linq;
-using VSRAD.Syntax.Options;
 
 namespace VSRAD.Syntax.FunctionList
 {
--- a/VSRAD.Syntax/Options/BaseOptionModel.cs
+++ b/VSRAD.Syntax/Options/BaseOptionModel.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.Serialization;
 using System.Runtime.Serialization.Formatters.Binary;
 using System.Threading.Tasks;
 using Microsoft;
@@ -106,7 +107,10 @@ namespace VSRAD.Syntax.Options
         {
             using (var stream = new MemoryStream())
             {
-                var formatter = new BinaryFormatter();
+                var formatter = new BinaryFormatter 
+                { 
+                    Binder = TypeOnlyBinder.Instance
+                };
                 formatter.Serialize(stream, value);
                 stream.Flush();
                 return Convert.ToBase64String(stream.ToArray());
@@ -119,7 +123,10 @@ namespace VSRAD.Syntax.Options
 
             using (var stream = new MemoryStream(b))
             {
-                var formatter = new BinaryFormatter();
+                var formatter = new BinaryFormatter 
+                { 
+                    Binder = TypeOnlyBinder.Instance
+                };
                 return formatter.Deserialize(stream);
             }
         }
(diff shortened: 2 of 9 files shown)


VSRAD.Syntax/Options/GeneralOptions.cs
```diff
@@ -0,0 +1,184 @@
...
+        [Description("by name")]
+        ByName = 3,
+        [Description("by name descending")]
+        ByNameDescending = 4,
+    }
+
+    public class GeneralOptions : BaseOptionModel<GeneralOptions>
+    {
+        private const string InstructionCollectionName = "RadeonAsmInstructionCollection";
+        private static readonly Regex fileExtensionRegular = new Regex(@"^\.\w+$");
```
johnny-keker, line 27: Code style consistency: rename private member to `_fileExtensionRegular`?

VSRAD.Syntax/FunctionList/FunctionListControl.xaml.cs
```diff
@@ -10,7 +10,6 @@
 using Task = System.Threading.Tasks.Task;
 using System.Threading;
 using System.Linq;
-using VSRAD.Syntax.Options;
 
 namespace VSRAD.Syntax.FunctionList
```
johnny-keker, line 14: In this file all private properties will be better to rename so they will start with `_`.

(for example, `private SortState SortState` is little bit confusing)

VSRAD.Syntax/Options/BaseOptionModel.cs
```diff
@@ -0,0 +1,154 @@
...
+            ShellSettingsManager manager = await _settingsManager.GetValueAsync();
+            WritableSettingsStore settingsStore = manager.GetWritableSettingsStore(SettingsScope.UserSettings);
+
+            if (!settingsStore.CollectionExists(CollectionName))
+                settingsStore.CreateCollection(CollectionName);
+
+            foreach (PropertyInfo property in GetOptionProperties())
+            {
+                string output = SerializeValue(property.GetValue(this));
+                settingsStore.SetString(CollectionName, property.Name, output);
```
johnny-keker, line 86: Code style consistency: it is better to use `var` when possible, or always use explicit type names

VSRAD.Syntax/Options/BaseOptionModel.cs
```diff
@@ -0,0 +1,154 @@
...
+                };
+                formatter.Serialize(stream, value);
+                stream.Flush();
+                return Convert.ToBase64String(stream.ToArray());
+            }
+        }
+
+        protected virtual object DeserializeValue(string value, Type type)
+        {
+            byte[] b = Convert.FromBase64String(value);
```
johnny-keker, line 111: Same as above

VSRAD.Syntax/Options/GeneralOptions.cs
```diff
@@ -0,0 +1,184 @@
...
+    public enum SortState
+    {
+        [Description("by line number")]
+        ByLine = 1,
+        [Description("by line number descending")]
+        ByLineDescending = 2,
+        [Description("by name")]
+        ByName = 3,
+        [Description("by name descending")]
+        ByNameDescending = 4,
```
johnny-keker, line 21: Explicit `int` values is unnecessary.

johnny-keker (changes requested):

VSRAD.Syntax/Options/GeneralOptions.cs
Slimakanzer: Thank you, fixed in 024da88

VSRAD.Syntax/Options/BaseOptionModel.cs
Slimakanzer: Thank you, fixed in 024da88

VSRAD.Syntax/Options/BaseOptionModel.cs
Slimakanzer: Fixed in 024da88

VSRAD.Syntax/Options/GeneralOptions.cs
Slimakanzer: Thank you, fixed in d95146a

VSRAD.Syntax/FunctionList/FunctionListControl.xaml.cs
Slimakanzer: This is a good comment, but not within the scope of this pull request.

johnny-keker (approved):
LGTM

PR #245, opened 2021-07-20
Combine instruction completions

Slimakanzer:
This pull request fixes multiple instances of instructions in the autocompletion list.

For example, if all instruction sets are selected (`instruction set selector -> All`).
Then completion list contained multiple **identical instructions** from different _Instruction Sets_.
image

This pull request fixes this issue:
image


johnny-keker (approved):
LGTM

PR #248, opened 2021-07-27
Fix stale analysis result on document relaod

Slimakanzer:
This PR fixes syntax highlighting "disappearing" when the file is reloaded. 
Original PR which is already in `dev` branch: #207 

This usually happens when the file is reloaded from disk but the content is unchanged (e.g. when using Disassemble/Preprocess actions from VSRAD.Package). In this case, the analysis result was not updated, which led to the disappearance of the syntax highlighting.


johnny-keker (approved):
Done some testing, LGTM

Kirpich30000 (approved):
lgtm

PR #254, opened 2021-07-29
Configs refactor

johnny-keker:
This PR provides changes from #188, #191 and fix NRE if old config file didn't had `TargetHosts` field from #201 but targeted on `dev-stable` branch.


VSRAD.Package/Options/ProjectOptions.cs
```diff
@@ -72,63 +81,99 @@ public void UpdateActiveProfile(ProfileOptions newProfile)
...
+        {
             try
             {
-                WriteAtomic(path, serializedOptions);
+                WriteAtomic(destPath, contents);
             }
             catch (UnauthorizedAccessException)
             {
-                DialogResult res = MessageBox.Show($"RAD Debug is unable to save configuration, because {path} is read-only. Make it writable?", "Warning", MessageBoxButtons.OKCancel, MessageBoxIcon.Warning);
+                if (File.ReadAllText(destPath) == contents) return; // We don't want to show the warning if config haven't changed
```
timlathy, line 160: I think it's a good idea to wrap `File.ReadAllText` in a `try-catch` block as well. No need to display anything in case of an exception, just treat it as if the config is changed.

A small example of how to crash VS with this:
1. Deny `Read` and `Write` permissions on the config file for your user account
2. Open Visual Studio, navigate to the profile editor, click `Apply` without changing anything
3. Boom

johnny-keker: Thanks! Resolved in 0642b6b

timlathy (approved):

PR #265, opened 2021-09-23
Action Names handling

### LINKED ISSUE #263: Duplicating Action names UB
Currently, we can create two Actions with the same name, which leads to UB for example in `Run Action` dialog or when assigning actions to the toolbar buttons.

We should either prohibit duplication of Action names or create a suffix system (which is prefered).

johnny-keker:
Resolves #263 

This PR improves validation of the Action names. Previously, it was possible to have two Actions with the same name, which in some cases (for example `RunAction`) could lead to the undefined behaviour and even to crash of the Visual Studio.

This PR fixes possible crashed and adds new features:

* Now new Actions added by a `+` button are provided with unique number suffix to prevent creating of two Actions with the same name:
image

* Added additional validation for renaming an Action, which ensures that new Action name is unique:
image


johnny-keker:
see #294

PR #270, opened 2021-10-01
Profile dropdown

### LINKED ISSUE #172: Add profile selector to RAD toolbar
It would be nice to have the same profile selector in the toolbar as we have in RAD Options dialog.

johnny-keker:
Resolves #172.

This PR adds `Debug Profile` dropdown to the `RAD Toolbar`.

image


timlathy (changes requested):
Overall LGTM, the dropdown behaves as expected in the following scenarios:
1) when using it to change the active profile,
2) when changing the active profile in the Options window,
3) when switching between projects and solutions.

However, there are some things that could be improved:
1. When there are no profiles for the current project, the dropdown still displays "Default". I think it'd be better to mirror the behavior of the hosts dropdown: if there are no profiles, nothing should be displayed.
2. The class naming is somewhat confusing: `ProfileDropdownCommand` vs `DebugProfileDropdownCommand`. I think the idea behind naming the old class `Profile...` was that it would manage both profiles and hosts. This makes more sense than it may initially seem because there's already an implicit dependency between the new and the existing class. Without `ProfileDropdownCommand.ProjectOptionsChanged`, `DebugProfileDropdownCommand` would not function correctly -- the dropdown would not update when the profile is changed in _Options_. I suggest merging the two classes, getting rid of `ProfileSelectorCommandSet` in the process. This should magically solve the first problem too :) given that `ProfileDropdownCommand` already handles the no profiles case.

johnny-keker:
> Overall LGTM, the dropdown behaves as expected in the following scenarios:
> 
> 1. when using it to change the active profile,
> 2. when changing the active profile in the Options window,
> 3. when switching between projects and solutions.
> 
> However, there are some things that could be improved:
> 
> 1. When there are no profiles for the current project, the dropdown still displays "Default". I think it'd be better to mirror the behavior of the hosts dropdown: if there are no profiles, nothing should be displayed.
> 2. The class naming is somewhat confusing: `ProfileDropdownCommand` vs `DebugProfileDropdownCommand`. I think the idea behind naming the old class `Profile...` was that it would manage both profiles and hosts. This makes more sense than it may initially seem because there's already an implicit dependency between the new and the existing class. Without `ProfileDropdownCommand.ProjectOptionsChanged`, `DebugProfileDropdownCommand` would not function correctly -- the dropdown would not update when the profile is changed in _Options_. I suggest merging the two classes, getting rid of `ProfileSelectorCommandSet` in the process. This should magically solve the first problem too :) given that `ProfileDropdownCommand` already handles the no profiles case.

Thanks, done in 7775f49

timlathy (approved):

PR #271, opened 2021-10-05
Slice visualizer [stable]

johnny-keker:
#108 but targeted on `dev-stable`.

Original description:

Addresses #71 

Done:
* Status string at the bottom with info about cell under mouse cursor: Group#, Column#, Value
* Right click context menu has Go to watch list which opens Visualizer window (if not opened), switches to a required group#, unhides that value (if hidden), scrolls to a corresponding value and selects it
* `Show columns` textbox for entering currently displayed columns
* `Watch` combobox contains only `System` and valid watches
* Each cell has the same color as a corresponding cell in Visualizer table
* Slice Visualizer button has an icon similar to Visualizer but with cells shaded like a heatmap
* Limited maximum column count in Slice and Debug Visualizers to 8192 to prevent undefined behavior

Not done yet:
* `Transposed view` feature


johnny-keker:
closed until further design discussion

PR #273, opened 2021-10-07
Improve handling of long-running/hanged actions [stable]

johnny-keker:
#194 but targeted on `dev-stable`.

Original description by @timlathy:

This PR allows the currently running action to be cancelled at any point in time. This is mostly useful when an `Execute` step starts a process that hangs or takes too long to finish. Previously, this behavior was available only for remote actions and had negative side-effects: _cancelling_ meant force-disconnecting from the server. This effectively terminated the action on the VS side but didn't stop the server from still running the command (e.g. waiting for the process to exit), which, coupled with the fact that commands from different clients cannot execute simultaneously, could cause the server to become unresponsive. To prevent this situation, a bidirectional ping mechanism is added. Currently it's only used by the server in the `Execute` command handler to detect when a client disconnects and terminate the process.

To preserve backward compatibility, the capability exchange (when the extension connects to the server) now includes extension capabilities as well. This way, the server knows if the client supports new behaviors (like pings) and can use fallbacks otherwise.

For local actions, cancellation is currently only supported by the process observer code. When a local `Execute` step hangs, cancelling the action will terminate the process. In future, cancellation can support can be added to other local steps as well, though I'm not aware of situations where they can take too long to complete.

In addition to that, this changes the behavior of the `Execute` step to show a confirmation dialog before terminating the process (with an option to wait for the process to exit). This requires new commands/responses:
* `ExecutionTimedOutResponse`, sent by the server when a timeout is reached, contains a list of processes that timed out (shown in the confirmation dialog);
* `ExecutionTimedOutActionCommand`, sent by the extension to indicate what action needs to be taken (terminate the process/wait);
* `ExecutionTerminatedResponse`, sent by the server in response to `ExecutionTimedOutActionCommand { TerminateProcesses = true }`, contains a list of processes that were terminated.

The server does not use these responses if the extension capabilities do not include `Base`, indicating a legacy extension version. In this case, the default behavior of terminating processes immediately is assumed, and instead of `ExecutionTerminatedResponse` the server sends `ExecutionCompleted { Status = TimedOut }`.

The old behavior can be enabled in the `Execute` step settings by setting `Timeout Behavior` to `Terminate`.


VSRAD.Package/Server/ActionRunner.cs
```diff
@@ -15,18 +16,29 @@
...
     {
         private readonly ICommunicationChannel _channel;
         private readonly SVsServiceProvider _serviceProvider;
+        private readonly IActionRunController _controller;
         private readonly Dictionary<string, DateTime> _initialTimestamps = new Dictionary<string, DateTime>();
         private readonly ReadOnlyCollection<string> _debugWatches;
         private readonly IProject _project;
 
-        public ActionRunner(ICommunicationChannel channel, SVsServiceProvider serviceProvider, ReadOnlyCollection<string> debugWatches, IProject project)
+        public ActionRunner(ICommunicationChannel channel, SVsServiceProvider serviceProvider, IActionRunController controller, ReadOnlyCollection<string> debugWatches, IProject project)
```
timlathy, line 37: `SVsServiceProvider serviceProvider` should be removed from the constructor parameters; `_serviceProvider` should be removed as well. This will probably require updating `ActionRunnerTests`.

VSRAD.Package/Server/ActionRunner.cs
```diff
@@ -254,6 +270,13 @@ private async Task<StepResult> DoCopySingleFileAsync(CopyFileStep step)
             return new StepResult(true, "", "");
         }
 
+        private async Task<StepResult> DoOpenInEditorAsync(OpenInEditorStep step)
```
timlathy, line 273: This method was moved to `IActionRunController.OpenFileInVsEditorAsync` and can be removed from `ActionRunner`.

VSRAD.DebugServer/SharedUtils/ProcessUtils.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    public static class ProcessUtils
+    {
+        public static bool IsParentOf(this Process parentProcess, Process otherProcess)
+        {
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                foreach (var line in File.ReadLines($"/proc/{otherProcess.Id}/status"))
+                {
+                    if (!line.StartsWith("PPid:", StringComparison.InvariantCultureIgnoreCase)) continue;
```
timlathy, line 39: Minor nitpick: according to [best practices](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings) it's more appropriate to use `OrdinalIgnoreCase` here. Relevant excerpts:
> * Use StringComparison.Ordinal or StringComparison.OrdinalIgnoreCase for comparisons as your safe default for culture-agnostic string matching.
> * Use comparisons with StringComparison.Ordinal or StringComparison.OrdinalIgnoreCase for better performance.
> * Do not use string operations based on StringComparison.InvariantCulture in most cases. One of the few exceptions is when you are persisting linguistically meaningful but culturally agnostic data.

timlathy (changes requested):

timlathy (approved):

PR #279, opened 2021-11-08
Asm2 lexer update

Slimakanzer:
This PR adds new lexer tokens for `asm2` syntax.

### Escape token
Added escape `\` token to fix identifier parsing which is joined with `\`.

Previously, **_arg1_** and **_arg2_** were not tokenized as identifiers (**_arg1\\_** and **_arg2\\_** were _UNKNOWN_ tokens). Because of what the parser did not work correctly in such a situation.  Now **_arg1\\_** is tokenized as 2 different tokens - _IDENTIFIER_, _ESCAPE_.

```
function (arg1\
         ,arg2\
         ,arg3)

end
```

### Native functions
Added new native functions: `float, floor, map, zip, head, tail, take, drop, rotate, replicate, reverse, cons, concat, list, s_list, v_list`

### Arrow operator
Added right arrow `->` operator used by lambdas.

### DIFF
--- a/VSRAD.Syntax/Core/Lexer/Asm2Lexer.cs
+++ b/VSRAD.Syntax/Core/Lexer/Asm2Lexer.cs
@@ -1,4 +1,4 @@
-﻿using Antlr4.Runtime;
+using Antlr4.Runtime;
 using Microsoft.VisualStudio.Text;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


johnny-keker (approved):
Done some testing.

* Operands with escape tokens are handled correctly;
* New native functions are highlighted;
* Arrow operator is highlighted.

LGTM

PR #280, opened 2021-11-08
Fix highlighting identifier concatinated with escape token

Slimakanzer:
Fixed identifier highlighting for such case (`value1`):

```
var value1 = 1
var value2 = 2
some_func(value1\
         ,value2)
```


johnny-keker (approved):
Escape token now is not included in the highlighted area when cursor is inside the operand.

LGTM

PR #287, opened 2021-12-10
Fix memory leaks [stable]

johnny-keker:
This PR is readding @timlathy's logic to prevent memory leaks on solution close.


johnny-keker (commented):
Checked with memory profiler - LGTM

PR #299, opened 2022-06-03
Environment Variables for Execute step

johnny-keker:
This PR adds a new textbox named `Environment Variables` to the `Execute` step properties. Specified environment variables will be set for the execution of the command.


johnny-keker:
Should be postponed until we port `Debug Server Capabilities` to current `master` branch

PR #301, opened 2022-07-07
Fix: Function List proper Open in Editor handling

johnny-keker:
This PR fixes incorrect handling of opening new file in editor by the `Function List`.

Steps to reproduce:
1.  Open any assembly source file in project, so `Function List` will update according to it's contents
2. Check both `Open in Editor: ...` checkboxes inside `RAD Options`
3. Create action that contains `Open in Editor` step with other assembly source file, that should trigger `Function List` to update
4. Run this action

You will see, that new pane is created for the source file from the `Open in Editor` step while active document tab should remain the same. However, `Function List` will be updated with contents of newly opened file, so it will be desynchonized with current active document, which is not expected.

With changes in this PR `Function List` won't update if newly opened document is not active, so in same case it will continue to show functions from original document.


Slimakanzer (approved):
LGTM!

PR #310, opened 2022-07-28
Name column resize master

### LINKED ISSUE #307: Watch name column resize
We want to manually resize Watch name column in `Visualizer`

johnny-keker:
Resolves #307 

This PR modifies the behavior of Name column of Debug Visualizer. Instead of auto-setting the width based on contents, now user can scale this column manually.

### DIFF
--- a/VSRAD.Package/DebugVisualizer/VisualizerTable.cs
+++ b/VSRAD.Package/DebugVisualizer/VisualizerTable.cs
@@ -83,6 +83,7 @@ namespace VSRAD.Package.DebugVisualizer
             _state = new TableState(this, columnWidth: 60, NameColumnIndex);
             _state.NameColumnScalingEnabled = true;
             SetupColumns();
+
             Debug.Assert(_state.DataColumnOffset == DataColumnOffset);
             Debug.Assert(_state.PhantomColumnIndex == PhantomColumnIndex);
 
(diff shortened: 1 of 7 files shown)


VSRAD.Package/DebugVisualizer/VisualizerTable.cs
```diff
@@ -248,7 +250,6 @@ private void SetupColumns()
                 HeaderText = "Name",
                 ReadOnly = false,
                 Frozen = true,
-                AutoSizeMode = DataGridViewAutoSizeColumnMode.AllCells,
```
timlathy, line 251: This disables the old behavior of the column automatically resizing to fit the data, even if `Options.VisualizerAppearance.ScaleNameColumn` is set to false. Is it intentional?

While the manual mode is nice to have, personally I find it less convenient compared to what we have now. Perhaps when `ScaleNameColumn` is disabled, we should keep the old behavior _and_ also autoresize the Name column when it's being edited, like this:

autoresize

Because if I recall correctly not having this behavior was one of the reasons why we wanted a manual scaling mode in the first place.

johnny-keker: Thank you! Done in 70b0b27

VSRAD.Package/DebugVisualizer/VisualizerTable.cs
```diff
@@ -248,12 +269,14 @@ private void SetupColumns()
...
                 SortMode = DataGridViewColumnSortMode.NotSortable
             });
             CreateMissingDataColumns(DataColumnCount);
         }
 
+        public void InitializeNameColumnWidth() =>
+            Columns[NameColumnIndex].Width = Columns[NameColumnIndex].GetPreferredWidth(DataGridViewAutoSizeColumnMode.AllCells, true);
```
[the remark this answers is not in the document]
johnny-keker: Thanks! Done in [9675959](https://github.com/vsrad/radeon-asm-tools/pull/310/commits/96759599cf4edb53b8e4c79e96aa8bdefa7db159)

timlathy (approved):

Issue #303, opened 2022-07-28
Function List displays incomplete functions names


johnny-keker:
`Function List` sometimes can fuse functions with different names. Needs to be investigated and reproduced.

timlathy:
It looks like the function list computes text width based on the visible items and does not adjust it on scroll or when the window is resized. Adding new labels or changing existing ones does recompute the width, though.

![fl_incomplete_names](https://user-images.githubusercontent.com/32609527/182403279-01898fa3-8ae2-4f98-831a-e1269a8df46e.gif)

PR #315, opened 2022-08-09
Fix: Function List proper Open in Editor handling [master]

### LINKED ISSUE #313: Confusing Function List behavior with Open in Editor: Preserve active document
Normally, the contents of Function List are updated when the editor tab selection is changed. However, when a document is open by the OpenInEditor step, and the "Open in Editor: Preserve active document" option is enabled, the tab focus does not change but the Function List is still updated, leading to confusing behavior where FL displays the contents for an inactive tab and moving the cursor in the active document does not refresh it:

FL

johnny-keker:
Resolves #313 

Added additional things so #301 would work for `master`.

### DIFF
--- a/VSRAD.Syntax/Core/DocumentFactory.cs
+++ b/VSRAD.Syntax/Core/DocumentFactory.cs
@@ -47,7 +47,7 @@ namespace VSRAD.Syntax.Core
             _dte = dte;
         }
 
-        public string GetActiveDocumentPath() => _dte.ActiveDocument.FullName;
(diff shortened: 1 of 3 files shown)


johnny-keker:
Resolves #313

Kirpich30000:
Hmmm..., there is #316 which is supposed to address #313
Could you please check which PR should be used to solve #313?

johnny-keker:
> Hmmm..., there is #316 which is supposed to address #313 Could you please check which PR should be used to solve #313?

I believe that #316 is supposed to resolve #303, while this PR resolves #313

Kirpich30000:
Ok, thanks for the clarification

PR #316, opened 2022-08-10
Added auto adjusting function names on resize

kikimych:
[#303 fix](https://github.com/vsrad/radeon-asm-tools/issues/303)


johnny-keker:
@kikimych are you sure that this changes resolve #303 on your local setup? I wasn't able to reproduce fix, maybe you can provide some steps to reproduce it? Thanks!

PR #317, opened 2022-08-11
FL displays incomplete function names fix

### LINKED ISSUE #303: Function List displays incomplete functions names

johnny-keker:
Resolves #303 

Edited hacky way to autosize Function List columns and disabled horizontal scrollbar, which fixes the incomplete function names bug.


johnny-keker:
Probably there is better way to fix that, so close for now.

PR #319, opened 2022-08-12
Visualizer nested list support

### LINKED ISSUE #304: Visualizer nested list support
No straightforward way to add nested lists to the watch list.

* One option is to put `Add to watches as array` in the context menu for each watch (then clicking on `bb[5]` -> `Add to watches as array` from 0 to 1 will insert `bb[5][0]` and `bb[5][1]`)

johnny-keker:
Resolves #304 

___

This PR adds `Add to watches as array` item in the context menu for each watch, then clicking on `bb[5]` -> `Add to watches as array` from 0 to 1 will insert `bb[5][0]` and `bb[5][1]`

image

### DIFF
--- a/VSRAD.Package/DebugVisualizer/VisualizerTable.cs
+++ b/VSRAD.Package/DebugVisualizer/VisualizerTable.cs
@@ -6,6 +6,7 @@ using System.Drawing;
 using System.Linq;
 using System.Windows.Forms;
 using VSRAD.Package.Options;
+using VSRAD.Package.ProjectSystem;
 using VSRAD.Package.Utils;
 
 namespace VSRAD.Package.DebugVisualizer
@@ -49,7 +50,7 @@ namespace VSRAD.Package.DebugVisualizer
 
         private readonly TableState _state;
 
-        public VisualizerTable(ProjectOptions options, FontAndColorProvider fontAndColor, GetValidWatches getValidWatches) : base()
+        public VisualizerTable(ProjectOptions options, FontAndColorProvider fontAndColor, GetValidWatches getValidWatches, IToolWindowIntegration toolWindowIntegration) : base()
         {
             _fontAndColor = fontAndColor;
             _computedStyling = new ComputedColumnStyling();
@@ -85,7 +86,7 @@ namespace VSRAD.Package.DebugVisualizer
 
             _ = new ContextMenus.ContextMenuController(this, new ContextMenus.IContextMenu[]
             {
-                new ContextMenus.TypeContextMenu(this, VariableTypeChanged, AvgprStateChanged, ProcessCopy, InsertSeparatorRow),
+                new ContextMenus.TypeContextMenu(this, VariableTypeChanged, AvgprStateChanged, ProcessCopy, InsertSeparatorRow, toolWindowIntegration),
                 new ContextMenus.CopyContextMenu(this, ProcessCopy),
                 new ContextMenus.SubgroupContextMenu(this, _state, options.VisualizerColumnStyling, () => options.DebuggerOptions.GroupSize)
             });
--- a/VSRAD.Package/DebugVisualizer/VisualizerTable.cs
+++ b/VSRAD.Package/DebugVisualizer/VisualizerTable.cs
@@ -87,7 +87,7 @@ namespace VSRAD.Package.DebugVisualizer
             _ = new ContextMenus.ContextMenuController(this, new ContextMenus.IContextMenu[]
             {
                 new ContextMenus.TypeContextMenu(this, VariableTypeChanged, AvgprStateChanged, ProcessCopy, InsertSeparatorRow,
-                    AddWatchRange),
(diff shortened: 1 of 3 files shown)


timlathy:
Behavior-wise, everything seems correct, style-wise -- and this is my opinion, not necessarily the optimal approach -- I would avoid passing `IToolWindowIntegration` to `TypeContextMenu`. It creates a coupling between unrelated subsystems (the project system integration and the GUI), not to mention that `IToolWindowIntegration` itself is a bit of a hack (essentially it's global state for tool windows) and in future should be refactored away.

One alternative I can think of is to:
1) Move [`AddWatch`](https://github.com/vsrad/radeon-asm-tools/blob/master/VSRAD.Package/DebugVisualizer/VisualizerControl.xaml.cs#L174) from `VisualizerControl` to `VisualizerTable`. Instead of doing `Watches.Clear(); Watches.AddRange(_table.GetCurrentWatchState());`, the table can just raise a `WatchStateChanged` event, whose handler would do the same. As a result, there will be less code duplication and a clearer separation of responsibilities.
2) `TypeContextMenu` can accept a delegate in the constructor that would add a watch from `name, offset, count`. This way, we don't even need to pass `ProjectOptions` to it: `VisualizerTable` will take care of building a range and adding it to watches.

johnny-keker:
> Behavior-wise, everything seems correct, style-wise -- and this is my opinion, not necessarily the optimal approach -- I would avoid passing `IToolWindowIntegration` to `TypeContextMenu`. It creates a coupling between unrelated subsystems (the project system integration and the GUI), not to mention that `IToolWindowIntegration` itself is a bit of a hack (essentially it's global state for tool windows) and in future should be refactored away.
> 
> One alternative I can think of is to:
> 
>     1. Move [`AddWatch`](https://github.com/vsrad/radeon-asm-tools/blob/master/VSRAD.Package/DebugVisualizer/VisualizerControl.xaml.cs#L174) from `VisualizerControl` to `VisualizerTable`. Instead of doing `Watches.Clear(); Watches.AddRange(_table.GetCurrentWatchState());`, the table can just raise a `WatchStateChanged` event, whose handler would do the same. As a result, there will be less code duplication and a clearer separation of responsibilities.
> 
>     2. `TypeContextMenu` can accept a delegate in the constructor that would add a watch from `name, offset, count`. This way, we don't even need to pass `ProjectOptions` to it: `VisualizerTable` will take care of building a range and adding it to watches.

Thanks for suggestion! Implemented in [df7491d](https://github.com/vsrad/radeon-asm-tools/pull/319/commits/df7491df2b178db17964c78b378982895194ccb0), is this ok?

timlathy:
Yes, looks good. Minor nitpicks: I would rename `AddWatchFromVisualizerTable` to something like `AddWatchRange`, same for `AddWatch(string name, int from, int to)`, since these functions do not insert watches but rather watch ranges, and you can use `RaiseWatchStateChanged` instead of `WatchStateChanged(GetCurrentWatchState())`.

johnny-keker:
> Yes, looks good. Minor nitpicks: I would rename `AddWatchFromVisualizerTable` to something like `AddWatchRange`, same for `AddWatch(string name, int from, int to)`, since these functions do not insert watches but rather watch ranges, and you can use `RaiseWatchStateChanged` instead of `WatchStateChanged(GetCurrentWatchState())`.

Thanks, done in [d4468a3](https://github.com/vsrad/radeon-asm-tools/pull/319/commits/d4468a3f5947363ccfd05d6ab25364eaaeca022e)

VSRAD.Package/DebugVisualizer/VisualizerTable.cs
```diff
@@ -85,7 +86,8 @@ public VisualizerTable(ProjectOptions options, FontAndColorProvider fontAndColor
 
             _ = new ContextMenus.ContextMenuController(this, new ContextMenus.IContextMenu[]
             {
-                new ContextMenus.TypeContextMenu(this, VariableTypeChanged, AvgprStateChanged, ProcessCopy, InsertSeparatorRow),
+                new ContextMenus.TypeContextMenu(this, VariableTypeChanged, AvgprStateChanged, ProcessCopy, InsertSeparatorRow,
+                    AddWatchRange),
```
timlathy, line 90: `addWatchRange: (name, from, to) => ArrayRange.FormatArrayRangeWatch(name, from, to, options.VisualizerOptions.MatchBracketsOnAddToWatches).ToList().ForEach(AddWatch))`

Rationale:
* Passing `MatchBracketsOnAddToWatches` to `FormatArrayRangeWatch`
* Less code duplication

johnny-keker: Thanks, done in [c27480e](https://github.com/vsrad/radeon-asm-tools/pull/319/commits/c27480e0a708f94783bf026d34f57786f753a6a1)

timlathy (approved):

PR #321, opened 2022-08-23
Function list displays incomplete function names

kikimych:
Fixes [#303](https://github.com/vsrad/radeon-asm-tools/issues/303)
Added explicit column width calculation and trimming for concatenated names


johnny-keker:
fixed by #322

PR #322, opened 2022-08-31
Incomplete function names fix

kikimych:
Fixes [#303 ](https://github.com/vsrad/radeon-asm-tools/issues/303).
Column sizes are explicitly calculated on the rendering stage.


johnny-keker (approved):
LGTM!

PR #328, opened 2022-09-07
[BugFix] GroupSize value for NDRange3D mode

Slimakanzer:
This PR fixes GroupSize value for NDRange3D mode.
Previously, dispatch parameters in visualizer did not work properly for 3D groups.

Dispatch parameter example:
```
grid_size (32, 1, 8)
group_size (32, 1, 8)
wave_size 32
```


johnny-keker (approved):
LGTM!

PR #329, opened 2022-09-07
Version handshake

### LINKED ISSUE #324: Extension-Server minimal version handshake
We want to provide mechanism of synchronization between extension and server to prevent undefined behavior on versions mismatch.

Most probable realization - client knows minimal server version that it needs, server knows minimal client version. On connection establishing they perform a handshake, if some condition is not satisfied debug session won't start and user will get a proper message - "please update extension / server to version ...".

___

_Note:_ That will require careful handling of minimal versions updates.

johnny-keker:
Resolves #324 

This PR provides new command - `GetMinimalExtensionVersion`. When establishing connection, extension sends this command and server returns minimal extension version that it needs as well as it's own version. Extension also knows minimal server version, so it compares acquired data and shows proper error message on version mismatch.

_Note:_ Any change in Server or Debugger that breaks compatibility should also update constants:

https://github.com/vsrad/radeon-asm-tools/blob/79b92545df50df7fc8e7f06835c9e406f9c99b6a/VSRAD.DebugServer/Server.cs#L19

https://github.com/vsrad/radeon-asm-tools/blob/79b92545df50df7fc8e7f06835c9e406f9c99b6a/VSRAD.Package/Constants.cs#L70

### DIFF
--- a/VSRAD.DebugServer/Dispatcher.cs
+++ b/VSRAD.DebugServer/Dispatcher.cs
@@ -16,6 +16,7 @@ namespace VSRAD.DebugServer
             PutFileCommand pf => new PutFileHandler(pf).RunAsync(),
             Deploy d => new DeployHandler(d, clientLog).RunAsync(),
             ListEnvironmentVariables lev => new ListEnvironmentVariablesHandler(lev).RunAsync(),
+            GetMinimalExtensionVersion gmex => new GetMinimalExtensionVersionHandler(gmex).RunAsync(),
             _ => throw new ArgumentException($"Unknown command type {command.GetType()}"),
         };
     }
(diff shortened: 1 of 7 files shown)


kikimych:
I think it's better not to add separate commands for this case. It's the internal details of the communication process and the user does not need to know about them. So, it's better not to show handshake in the interface.

We had discussion with @Kirpich30000 and decided that it's better to process communication this way:

1. client sends its version
2. server sends its version
3. server checks client's version for compatibility and sends result to client
4. client checks server version for compatibility and sends result to server
5. if either client or server can not proceed connection closes

Also I think this process can be done without construction of any classes at least now. Version itself is simple String and does not need complex logic for serialization/deserialization

Regards, Alexey

johnny-keker:
Closing this PR since we decided to implement this feature other way.

PR #330, opened 2022-09-10
NDRange3D refactor

johnny-keker:
This PR provides refactoring for `NDRange3D` feature.

Changes:

* Removed `NDRange3D` from dispatch parameters
Looks like we did not had a good reason for it to be there, as the only real use of that was caused by setting `DimY` and `DimZ` to 0 in `BreakStateDispatchParameters` instead of 1, which is incorrect in the first place and also refactored.
* That means that calculation of `GroupSize` and `NGroups` no longer depends on the state of `NDRange3D` as 1-dimentional indexing is equivalent to 3-dimentional when `DimY` and `DimZ` are set to 1's.
* As a pleasant bonus - now 3D and 1D group coordinates are converted vice versa when user toggles `3D NDRange` (see attached gif)

group_coords

### DIFF
--- a/VSRAD.Package/DebugVisualizer/GroupIndexSelector.cs
+++ b/VSRAD.Package/DebugVisualizer/GroupIndexSelector.cs
@@ -50,7 +50,7 @@ namespace VSRAD.Package.DebugVisualizer
         public uint DimZ { get => _dimZ; set { SetField(ref _dimZ, value); RaisePropertyChanged(nameof(MaximumZ)); } }
 
         // OneWay bindings in XAML do not work on these properties for some reason, hence the empty setters
-        public uint MaximumX { get => _projectOptions.VisualizerOptions.NDRange3D ? DimX - 1 : uint.MaxValue; set { } }
+        public uint MaximumX { get => _projectOptions.VisualizerOptions.NDRange3D ? DimX - 1 : DimX * DimY * DimZ - 1; set { } }
         public uint MaximumY { get => DimY - 1; set { } }
         public uint MaximumZ { get => DimZ - 1; set { } }
 
@@ -78,7 +78,21 @@ namespace VSRAD.Package.DebugVisualizer
                         SetOptionsFromDispatchParams();
                     break;
                 case nameof(Options.VisualizerOptions.NDRange3D):
-                    RaisePropertyChanged(nameof(MaximumX));
+                    if (!_projectOptions.VisualizerOptions.NDRange3D)
+                    {
+                        RaisePropertyChanged(nameof(MaximumX));
+                        X = X + Y * DimX + Z * DimX * DimY;
+                        Y = 1;
+                        Z = 1;
+                    }
+                    else
+                    {
+                        var x = X;
+                        X = X % DimX;
+                        Y = x % (DimX * DimY) / DimZ;
+                        Z = x / DimX / DimY;
+                        RaisePropertyChanged(nameof(MaximumX));
+                    }
                     if (_updateOptions) Update();
                     break;
                 case nameof(Options.DebuggerOptions.WaveSize):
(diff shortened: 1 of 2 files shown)


VSRAD.Package/DebugVisualizer/GroupIndexSelector.cs
```diff
@@ -78,7 +78,21 @@ private void OptionsChanged(object sender, PropertyChangedEventArgs e)
                         SetOptionsFromDispatchParams();
                     break;
                 case nameof(Options.VisualizerOptions.NDRange3D):
-                    RaisePropertyChanged(nameof(MaximumX));
+                    if (!_projectOptions.VisualizerOptions.NDRange3D)
+                    {
+                        RaisePropertyChanged(nameof(MaximumX));
```
timlathy, line 83: 1. I don't remember why this line (`RaisePropertyChanged(...)`) was in the original code. It's not clear what it does and why it is needed, could you check what happens when you remove it?
2. If there's a reason for having this call before setting `X`, `Y`, `Z` with NDRange disabled and after setting `X`, `Y`, `Z` with NDRange enabled, then this should be reflected in code comments, otherwise the calling order should be identical (and this line can be moved outside of `if-else`).

VSRAD.Package/Server/BreakStateDispatchParameters.cs
```diff
@@ -31,9 +30,8 @@ private BreakStateDispatchParameters(uint waveSize, uint gridX, uint gridY, uint
             GridSizeY = Math.Max(1, gridY);
             GridSizeZ = Math.Max(1, gridZ);
             DimX = gridX / GroupSizeX;
-            DimY = gridY > 1 ? gridY / GroupSizeY : 0;
-            DimZ = gridZ > 1 ? gridZ / GroupSizeZ : 0;
-            NDRange3D = gridY > 1 || gridZ > 1;
+            DimY = gridY / GroupSizeY;
+            DimZ = gridZ / GroupSizeZ;
```
Slimakanzer, line 34: 1. I don't remember why `Math.Max` is needed, but I would change `DimX = gridX / GroupSizeX` to `DimX = GroupSizeX / GroupSizeX`. Same for `DimY`, `DimZ`.
2. Is there any reason for having `Math.Max(1, ...)`?

Slimakanzer (changes requested):
I suspect the problem is much more complicated than we thought. We need to carefully review all lines of code that use `VisualizerOptions.NDRange3D`.  In the following example, it's not clear why 1D `groupIdx` is used even with 3D Range. 

https://github.com/vsrad/radeon-asm-tools/blob/master/VSRAD.Package/DebugVisualizer/GroupIndexSelector.cs#L123-L140

VSRAD.Package/Server/BreakStateDispatchParameters.cs
timlathy: Regarding `Math.Max`, I suspect it's used in the constructor to enforce the invariant that `SizeX/Y/Z >= 1`. Which is not that bad of an idea, considering that we assume it holds true in other parts of the code. I wonder, though, if it would be better to throw an exception instead of silently swallowing invalid values.

VSRAD.Package/DebugVisualizer/GroupIndexSelector.cs
```diff
@@ -78,7 +78,21 @@ private void OptionsChanged(object sender, PropertyChangedEventArgs e)
...
+                        Z = 1;
+                    }
+                    else
+                    {
+                        var x = X;
+                        X = X % DimX;
+                        Y = x % (DimX * DimY) / DimZ;
+                        Z = x / DimX / DimY;
+                        RaisePropertyChanged(nameof(MaximumX));
+                    }
```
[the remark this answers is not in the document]
johnny-keker: This chunk of code is actually responsible for conversion 1D group index to 3D, and vice versa. So naturally, if we turned on 1D indexation, we need to do 3D->1D conversion, in other case - 1D->3D

PR #331, opened 2022-09-12
Error List multiple locations errors fix

### LINKED ISSUE #326: Error messages with multiple locations are missing from Error List
Note: this applies to script errors only (`ScriptErrorRegex`).

Example error message for tests:

```
*W,undefined: undefined reference to 'printf' (C:\Absolute\Path\source.c:3, C:\Absolute\Path\source.c:5, C:\Absolute\Path\include.h:10)
```

This message should generate three separate entries in Error List, one for each location.

johnny-keker:
Resolves #326 

This PR fixes the parsing of error messages with multiple locations for Error list and adds a corresponding test.

For more info refer to #326

### DIFF
--- a/VSRAD.Package/BuildTools/Errors/Parser.cs
+++ b/VSRAD.Package/BuildTools/Errors/Parser.cs
@@ -20,11 +20,20 @@ namespace VSRAD.Package.BuildTools.Errors
                     string line;
                     while ((line = reader.ReadLine()) != null)
                     {
-                        var message = ParseScriptMessage(line) ?? ParseKeywordMessage(line) ?? ParseClangMessage(line);
-                        if (message != null)
+                        var message = ParseKeywordMessage(line) ?? ParseClangMessage(line);
+                        var scriptMessages = ParseScriptMessage(line);
+                        if (message != null || scriptMessages != null)
                         {
-                            lastMessage = message;
-                            messages.Add(message);
+                            if (message != null)
+                            {
+                                lastMessage = message;
+                                messages.Add(message);
+                            }
+                            else
+                            {
+                                lastMessage = scriptMessages[scriptMessages.Count - 1];
+                                messages.AddRange(scriptMessages);
+                            }
                         }
                         else if (lastMessage != null)
                         {
@@ -54,12 +63,15 @@ namespace VSRAD.Package.BuildTools.Errors
         }
 
         private static readonly Regex ScriptErrorRegex = new Regex(
-            @"\*(?<kind>[EW]),(?<code>[^:(]+)(?>\s\((?<file>.+):(?<line>\d+)\))?:\s(?<text>.+)", RegexOptions.Compiled);
+            @"\*(?<kind>[EW]),(?<code>[^:(]+)(?>\s\((?<file>.+)\))?:\s(?<text>.+)", RegexOptions.Compiled);
 
-        private static readonly Regex ScriptErrorLocationInTextRegex = new Regex(
-            @"(?<text>.+)\s\((?<file>.+):(?<line>\d+)\)", RegexOptions.Compiled);
+        private static readonly Regex ScriptErrorTextRegex = new Regex(
+            @"(?<text>.+)\s\((?<file>.+)\)", RegexOptions.Compiled);
 
-        private static Message ParseScriptMessage(string header)
+        private static readonly Regex ScriptErrorLocationsRegex = new Regex(
+            @"(?<file>.+):(?<line>\d+)", RegexOptions.Compiled);
+
+        private static List<Message> ParseScriptMessage(string header)
         {
             var match = ScriptErrorRegex.Match(header);
             if (!match.Success) return null;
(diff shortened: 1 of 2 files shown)


VSRAD.Package/BuildTools/Errors/Parser.cs
```diff
@@ -20,11 +20,20 @@ public static ICollection<Message> ParseStderr(IEnumerable<string> outputs)
...
+                                lastMessage = message;
+                                messages.Add(message);
+                            }
+                            else
+                            {
+                                lastMessage = scriptMessages[scriptMessages.Count - 1];
+                                messages.AddRange(scriptMessages);
+                            }
                         }
                         else if (lastMessage != null)
```
timlathy, line 38: `lastMessage` is redundant I think, we can get rid of it if we change this condition to something like `if (messages.Count > 0) { messages.Last().Text += ... }`

VSRAD.Package/BuildTools/Errors/Parser.cs
```diff
@@ -20,11 +20,20 @@ public static ICollection<Message> ParseStderr(IEnumerable<string> outputs)
                     string line;
                     while ((line = reader.ReadLine()) != null)
                     {
-                        var message = ParseScriptMessage(line) ?? ParseKeywordMessage(line) ?? ParseClangMessage(line);
-                        if (message != null)
+                        var message = ParseKeywordMessage(line) ?? ParseClangMessage(line);
+                        var scriptMessages = ParseScriptMessage(line);
+                        if (message != null || scriptMessages != null)
```
timlathy, line 25: This condition is not necessary since  `message != null` and `scriptMessages != null` are handled differently.

VSRAD.Package/BuildTools/Errors/Parser.cs
```diff
@@ -54,33 +63,47 @@ private static Message ParseClangMessage(string header)
...
-        private static readonly Regex ScriptErrorLocationInTextRegex = new Regex(
-            @"(?<text>.+)\s\((?<file>.+):(?<line>\d+)\)", RegexOptions.Compiled);
+        private static readonly Regex ScriptErrorTextRegex = new Regex(
+            @"(?<text>.+)\s\((?<file>.+)\)", RegexOptions.Compiled);
 
-        private static Message ParseScriptMessage(string header)
+        private static readonly Regex ScriptErrorLocationsRegex = new Regex(
+            @"(?<file>.+):(?<line>\d+)", RegexOptions.Compiled);
+
+        private static List<Message> ParseScriptMessage(string header)
```
timlathy, line 74: 1. I'd change the function name to ParseScriptMessage**s** as it returns a list now.
2. I'm not sure if it's a good idea to return `null` from a function that returns a list. I would expect to get an empty list if there are no messages found in the string.
3. I wonder if making all Parse functions return a list would make the code in `ParseStderr` (specifically the fact that regexes are applied in a specific order and only one can match a given line) clearer. I'm not completely sure it's worth it though...

VSRAD.Package/BuildTools/Errors/Parser.cs
```diff
@@ -54,33 +63,47 @@ private static Message ParseClangMessage(string header)
         }
 
         private static readonly Regex ScriptErrorRegex = new Regex(
-            @"\*(?<kind>[EW]),(?<code>[^:(]+)(?>\s\((?<file>.+):(?<line>\d+)\))?:\s(?<text>.+)", RegexOptions.Compiled);
+            @"\*(?<kind>[EW]),(?<code>[^:(]+)(?>\s\((?<file>.+)\))?:\s(?<text>.+)", RegexOptions.Compiled);
 
-        private static readonly Regex ScriptErrorLocationInTextRegex = new Regex(
-            @"(?<text>.+)\s\((?<file>.+):(?<line>\d+)\)", RegexOptions.Compiled);
+        private static readonly Regex ScriptErrorTextRegex = new Regex(
+            @"(?<text>.+)\s\((?<file>.+)\)", RegexOptions.Compiled);
```
Slimakanzer, line 69: Not clear naming for `file` group. It can be file with location or it can be simple error text because error like `W,undefined: undefined reference to 'printf' (some error text)` will also match the regex.

VSRAD.Package/BuildTools/Errors/Parser.cs
```diff
@@ -54,33 +63,47 @@ private static Message ParseClangMessage(string header)
...
-                message.SourceFile = match.Groups["file"].Value;
-                message.Line = int.Parse(match.Groups["line"].Value);
+                foreach (var source in match.Groups["file"].Value.Split(','))
+                {
+                    var message = new Message { Kind = kind };
+                    var locationMatch = ScriptErrorLocationsRegex.Match(source);
+                    message.SourceFile = locationMatch.Groups["file"].Value.Trim();
+                    message.Line = int.Parse(locationMatch.Groups["line"].Value);
+                    message.Text = code + ": " + match.Groups["text"].Value;
+                    messages.Add(message);
```
Slimakanzer, line 98: This code will throw an error if `source` doesn't match `ScriptErrorLocationsRegex`. Example: `W,undefined: undefined reference to 'printf' (some error text)`. I would recommend you get rid of `Split` and handle commas using regular expression like `(?<file>[^,]+):(?<line>\d+)`.

timlathy:
Good news: with the latest changes to our tooling, messages like `*W,code: text (C:\Absolute\Path\source.c:3, C:\Absolute\Path\source.c:5, C:\Absolute\Path\include.h:10)` will no longer be generated, so there's no need to handle this format.

I would suggest keeping most of the changes in this PR with the exception of the code that handles locations after error text (it can be safely removed). I would also prefer to go back to the previous signature of Parse... methods (have them return Message or null). I think it's a good style to keep functions pure by avoiding in-out parameters where it's possible.

johnny-keker:
> Good news: with the latest changes to our tooling, messages like `*W,code: text (C:\Absolute\Path\source.c:3, C:\Absolute\Path\source.c:5, C:\Absolute\Path\include.h:10)` will no longer be generated, so there's no need to handle this format.
> 
> I would suggest keeping most of the changes in this PR with the exception of the code that handles locations after error text (it can be safely removed). I would also prefer to go back to the previous signature of Parse... methods (have them return Message or null). I think it's a good style to keep functions pure by avoiding in-out parameters where it's possible.

Thanks! Is https://github.com/vsrad/radeon-asm-tools/pull/331/commits/9fd8cb261da14a3095dad0908931131f32f29002 works for you?

PR #333, opened 2022-09-23
Added version checking in handshake during connection to debug server

kikimych:
Resolves [#324 ](url)

Added version handshake checking

### DIFF
--- a/VSRAD.DebugServer/Server.cs
+++ b/VSRAD.DebugServer/Server.cs
@@ -4,7 +4,9 @@ using System.Net.Sockets;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
-
+using System.IO;
+using System.Text;
+using VSRAD.DebugServer.IPC.Commands;
 namespace VSRAD.DebugServer
 {
     public sealed class Server
@@ -12,6 +14,8 @@ namespace VSRAD.DebugServer
         private readonly SemaphoreSlim _commandExecutionLock = new SemaphoreSlim(1, 1);
         private readonly TcpListener _listener;
         private readonly bool _verboseLogging;
+        private static Version _serverVersion = typeof(Server).Assembly.GetName().Version;
+        private static Version _minimalAcceptedClientVersion = new Version("2021.12.8");
 
         const uint ENABLE_QUICK_EDIT = 0x0040;
         const int STD_INPUT_HANDLE = -10;
@@ -49,24 +53,81 @@ namespace VSRAD.DebugServer
             while (true)
             {
                 var client = await _listener.AcceptTcpClientAsync().ConfigureAwait(false);
-                ClientConnected(client, clientsCount);
+                TcpClientConnected(client, clientsCount);
                 clientsCount++;
             }
         }
 
-        private void ClientConnected(TcpClient tcpClient, uint clientId)
+        private void TcpClientConnected(TcpClient tcpClient, uint clientId)
         {
             var networkClient = new NetworkClient(tcpClient, clientId);
             var clientLog = new ClientLogger(clientId, _verboseLogging);
+            if (!Task.Run(() => TryProcessServerHandshake(tcpClient, clientLog)).Result)
+            {
+                clientLog.HandshakeFailed(networkClient.EndPoint);
+                return;
+            }
             clientLog.ConnectionEstablished(networkClient.EndPoint);
             Task.Run(() => BeginClientLoopAsync(networkClient, clientLog));
+        }
+
+        private async Task<bool> TryProcessServerHandshake(TcpClient client, ClientLogger clientLog)
+        {
+            try
+            {
+                var writer = new StreamWriter(client.GetStream(), Encoding.UTF8) { AutoFlush = true };
+                var reader = new StreamReader(client.GetStream(), Encoding.UTF8);
+
+                // Send server version to client
+                //
+                await writer.WriteLineAsync(_serverVersion.ToString()).ConfigureAwait(false);
+
+                // Obtain client version
+                //
+                String clientResponse = await reader.ReadLineAsync().ConfigureAwait(false);
+
+                if (!Version.TryParse(clientResponse, out var clientVersion))
+                {
+                    clientLog.ParseVersionError(clientResponse);
+                    // Inform client that server declines client's version
+                    //
+                    await writer.WriteLineAsync(HandShakeStatus.ServerNotAccepted.ToString()).ConfigureAwait(false);
+                    return false;
+                }
+
+                if (clientVersion.CompareTo(_minimalAcceptedClientVersion) < 0)
+                {
+                    clientLog.InvalidVersion(clientVersion.ToString(), _minimalAcceptedClientVersion.ToString());
+                    // Inform client that server declines client's version
+                    //
+                    await writer.WriteLineAsync(HandShakeStatus.ServerNotAccepted.ToString()).ConfigureAwait(false);
+                    return false;
+                }
+
+                // Inform client that server accepts client's version
+                //
+                await writer.WriteLineAsync(HandShakeStatus.ServerAccepted.ToString()).ConfigureAwait(false);
+
+                // Check if client accepts server version
+                //
+                if (await reader.ReadLineAsync() != HandShakeStatus.ClientAccepted.ToString())
+                {
+                    clientLog.ClientRejectedServerVersion(_serverVersion.ToString(), clientVersion.ToString());
+                    return false;
+                }
+            } catch (Exception)
+            {
+                clientLog.ConnectionTimeoutOnHandShake();
+                return false;
+            }
+            return true;
         }
 
         private async Task BeginClientLoopAsync(NetworkClient client, ClientLogger clientLog)
         {
             while (true)
             {
-                bool lockAcquired = false;
+                var lockAcquired = false;
                 try
                 {
                     var command = await client.ReceiveCommandAsync().ConfigureAwait(false);
--- a/VSRAD.Package/Server/CommunicationChannel.cs
+++ b/VSRAD.Package/Server/CommunicationChannel.cs
@@ -12,6 +12,10 @@ using VSRAD.DebugServer.IPC.Responses;
 using VSRAD.Package.Options;
 using VSRAD.Package.ProjectSystem;
 using Task = System.Threading.Tasks.Task;
+using System.IO;
+using System.Text;
+using System.Text.RegularExpressions;
+using VSRAD.DebugServer.IPC.Commands;
 
 namespace VSRAD.Package.Server
 {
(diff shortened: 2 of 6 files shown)


johnny-keker (changes requested):
I am not sure about message we get when extension version is below minimal required on server (see attached image). We get required _server_ version in message, however I believe we should display required _extension_ version in the context of the message?

Also, little nitpick - would be nice to modify version format in the error message so it would match our style of versioning

prob

kikimych:
> 

By default client have to update to server's version. There is guarantee that client and server are compatible on the same version's

johnny-keker:
> > 
> 
> By default client have to update to server's version. There is guarantee that client and server are compatible on the same version's

I am questioning this error message wording.

https://github.com/vsrad/radeon-asm-tools/blob/acac07beb8f69152b06c799ceb9ea6d9ca4b66ac/VSRAD.Package/Server/CommunicationChannel.cs#L53

"This _extension_ is out of date... Please update it to the `{serverVersion}`..."

By _it_ we mean server, or is it extension?

From the context of the error looks like we mean extension?

VSRAD.DebugServer/Server.cs
```diff
@@ -12,6 +14,8 @@ public sealed class Server
         private readonly SemaphoreSlim _commandExecutionLock = new SemaphoreSlim(1, 1);
         private readonly TcpListener _listener;
         private readonly bool _verboseLogging;
+        private static Version _serverVersion = typeof(Server).Assembly.GetName().Version;
+        private static Version _minimalAcceptedClientVersion = new Version("2021.12.8");
```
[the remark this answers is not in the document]
kikimych: There is incompatibility of protocols between old server and new client and vice versa. It can't be solved seamlessly.

kikimych:
> > > 
> > 
> > 
> > By default client have to update to server's version. There is guarantee that client and server are compatible on the same version's
> 
> I am questioning this error message wording.
> 
> https://github.com/vsrad/radeon-asm-tools/blob/acac07beb8f69152b06c799ceb9ea6d9ca4b66ac/VSRAD.Package/Server/CommunicationChannel.cs#L53
> 
> "This _extension_ is out of date... Please update it to the `{serverVersion}`..."
> 
> By _it_ we mean server, or is it extension?
> 
> From the context of the error looks like we mean extension?

fixed

johnny-keker:
> > > > 
> > > 
> > > 
> > > By default client have to update to server's version. There is guarantee that client and server are compatible on the same version's
> > 
> > 
> > I am questioning this error message wording.
> > https://github.com/vsrad/radeon-asm-tools/blob/acac07beb8f69152b06c799ceb9ea6d9ca4b66ac/VSRAD.Package/Server/CommunicationChannel.cs#L53
> > 
> > "This _extension_ is out of date... Please update it to the `{serverVersion}`..."
> > By _it_ we mean server, or is it extension?
> > From the context of the error looks like we mean extension?
> 
> fixed

`Please update extension to the {serverVersion}...`

I may be totally wrong, but I fail to understand why would I want to update _extension_ to the latest _server_ version? We can't guarantee that latest extension's and server's version are the same.

However, what does `serverVersion` mean here? Is it actual minimal server version, or minimal extension version that we got from server? In second case, it is misunderstanding and I believe that this variable should be renamed to something like `extensionVersion`.

VSRAD.Package/Server/CommunicationChannel.cs
```diff
@@ -81,11 +107,56 @@ public CommunicationChannel(SVsServiceProvider provider, IProject project)
...
+            await writer.WriteLineAsync(_extensionVersion.ToString()).ConfigureAwait(false);
+            // Obtain server version
+            //
+            String serverResponse = await reader.ReadLineAsync().ConfigureAwait(false);
+            Version serverVersion = null;
+            if (!Version.TryParse(serverResponse, out serverVersion))
+            {
+                // Inform server that client declines serve's version
+                //
+                await writer.WriteLineAsync(HandShakeStatus.client_not_accepted.ToString()).ConfigureAwait(false);
```
johnny-keker, line 135: Here and below: case of `HandShakeStatus` changed only in definition, but not in usages

kikimych:
> > > > > 
> > > > 
> > > > 
> > > > By default client have to update to server's version. There is guarantee that client and server are compatible on the same version's
> > > 
> > > 
> > > I am questioning this error message wording.
> > > https://github.com/vsrad/radeon-asm-tools/blob/acac07beb8f69152b06c799ceb9ea6d9ca4b66ac/VSRAD.Package/Server/CommunicationChannel.cs#L53
> > > 
> > > "This _extension_ is out of date... Please update it to the `{serverVersion}`..."
> > > By _it_ we mean server, or is it extension?
> > > From the context of the error looks like we mean extension?
> > 
> > 
> > fixed
> 
> `Please update extension to the {serverVersion}...`
> 
> I may be totally wrong, but I fail to understand why would I want to update _extension_ to the latest _server_ version? We can't guarantee that latest extension's and server's version are the same.
> 
> However, what does `serverVersion` mean here? Is it actual minimal server version, or minimal extension version that we got from server? In second case, it is misunderstanding and I believe that this variable should be renamed to something like `extensionVersion`.

Extension and server are on same repository at this moment. So, it's possible to hold this invariant.

johnny-keker (changes requested):
I would suggest move `HandShakeStatus` structure in some file in `IPC` directory on server side, because these files are automatically symlinked into `VSRAD.Package.Server.IPC` and we would not need to duplicate this structure on extension side

johnny-keker:
Also, I strongly recommend cast enum entries to `byte` before calling `ToString()` and sending them, because current serialization not only not effective, but also can cause some problems. Wasted a lot of time trying to figure out why handshake broke after simple changing case of enum entries in code, until realized that I actually need to change it in two different places _and also_ rebuild Debug Server

VSRAD.DebugServer/Server.cs
```diff
@@ -31,6 +35,14 @@ public Server(IPAddress ip, int port, bool verboseLogging = false)
             _verboseLogging = verboseLogging;
         }
 
+        public enum HandShakeStatus
+        {
+            CLIENT_ACCEPTED,
+            CLIENT_NOT_ACCEPTED,
+            SERVER_ACCEPTED,
+            SERVER_NOT_ACCEPTED
```
johnny-keker, line 43: To match our code style I would go with `ClientAccepted` etc

VSRAD.DebugServer/Server.cs
```diff
@@ -49,19 +61,77 @@ public async Task LoopAsync()
...
             clientLog.ConnectionEstablished(networkClient.EndPoint);
             Task.Run(() => BeginClientLoopAsync(networkClient, clientLog));
         }
 
+        private async Task<bool> TryProcessServerHandshake(TcpClient client, ClientLogger clientLog)
+        {
+            try
+            {
+                StreamWriter writer = new StreamWriter(client.GetStream(), Encoding.UTF8) { AutoFlush = true };
+                StreamReader reader = new StreamReader(client.GetStream(), Encoding.UTF8);
```
johnny-keker, line 87: In this project (at least in Package and Server, not so sure about Syntax) we tend not to use explicit types where possible, so here and below (`clientResponse`) I would definitely go with `var` keyword

VSRAD.DebugServer/Server.cs
```diff
@@ -49,19 +61,77 @@ public async Task LoopAsync()
...
+                // Send server version to client
+                //
+                await writer.WriteLineAsync(_serverVersion.ToString()).ConfigureAwait(false);
+
+                // Obtain client version
+                //
+                String clientResponse = await reader.ReadLineAsync().ConfigureAwait(false);
+
+                Version clientVersion = null;
+                if (!Version.TryParse(clientResponse, out clientVersion))
```
johnny-keker, line 98: Also, to better match project codestyle (and my personal preference, this is not necessary), I would go with single line here:

```C#
if (!Version.TryParse(clientResponse, out var clientVersion))
```

VSRAD.Package/Server/CommunicationChannel.cs
```diff
@@ -37,21 +40,43 @@ public sealed class ConnectionRefusedException : System.IO.IOException
         { }
     }
 
+    public sealed class UnsupportedServerVersionException : System.IO.IOException
```
johnny-keker, line 43: Here and below, we can cut `System.IO.`

VSRAD.Package/Server/CommunicationChannel.cs
```diff
@@ -81,11 +107,56 @@ public CommunicationChannel(SVsServiceProvider provider, IProject project)
...
+            var match = _extensionVersionRegex.Match(typeof(IProject).Assembly.CodeBase);
+            if (!match.Success)
+                throw new Exception("Error while getting current extension version.");
+            _extensionVersion = new Version(match.Groups["version"].Value);
         }
 
         public Task<T> SendWithReplyAsync<T>(ICommand command) where T : IResponse =>
             SendWithReplyAsync<T>(command, tryReconnect: true);
 
+        private async Task<bool> TryProcessClientHandshake(TcpClient client)
```
johnny-keker, line 119: This method doesn't need to return `bool`, since result is not checked and it returns only `true` or exception. I would go with simple `Task` here. Also, Visual Studio gives warning, since all async methods are expected to have an `Async` suffix in name, so I would suggest renaming this method to `TryProcessClientHandshakeAsync`

VSRAD.Package/Server/CommunicationChannel.cs
```diff
@@ -81,11 +107,56 @@ public CommunicationChannel(SVsServiceProvider provider, IProject project)
...
+        {
+            StreamWriter writer = new StreamWriter(client.GetStream(), Encoding.UTF8) { AutoFlush = true };
+            StreamReader reader = new StreamReader(client.GetStream(), Encoding.UTF8);
```
johnny-keker, line 122: Same as in `Server.cs`

johnny-keker (changes requested):

VSRAD.Package/Server/CommunicationChannel.cs
kikimych: done

VSRAD.Package/Server/CommunicationChannel.cs
kikimych: done

VSRAD.Package/Server/CommunicationChannel.cs
kikimych: done

VSRAD.DebugServer/Server.cs
kikimych: done

VSRAD.DebugServer/Server.cs
kikimych: done

VSRAD.DebugServer/Server.cs
kikimych: done

johnny-keker (approved):
LGTM!

PR #334, opened 2022-10-31
Aliases for Target Hosts

### LINKED ISSUE #332: Add IP and Port aliases for Remote host dropdown
```
> ALIAS (<ip>:port)
> ALIAS (<ip>:port)
...
```

johnny-keker:
Resolves #332 

Reworked Target Hosts edition dialog (see image). Now host and port are separate fields (latter will be set to default `9339` when new host is added), and also we have optional `Alias` field. If set to non-empty, this host will be listed by its alias in dropdown.

User still can manually enter new host to the hosts dropdown in format `ip:port`. User input is carefully handled - if there is such host, but wit alias, alias will be used. If user just enters known alias, corresponding host will be used.

The last, but not least, this PR contains a work on refactoring the logic of storing and using current target host. Before, there were a lot of duplication of essentially same info (string `ip:port`, string `ip` and ushort `port`, `HostItem`, `ServerConnectionOptions`...), now it is standardized and much easier to understand and extend (f.e. add an alias). Also, `RemoteMachine` and `Port` are gone from project config, since this info is already stored in `TargetHosts`, which works like MRU collection.

image

### DIFF
--- a/VSRAD.Package/Commands/ProfileDropdownCommand.cs
+++ b/VSRAD.Package/Commands/ProfileDropdownCommand.cs
@@ -11,6 +11,7 @@ using System.Windows;
 using VSRAD.Package.Options;
 using VSRAD.Package.ProjectSystem;
 using VSRAD.Package.ProjectSystem.Profiles;
+using VSRAD.Package.Utils;
 
 namespace VSRAD.Package.Commands
 {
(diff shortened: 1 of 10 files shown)


kikimych:
Could you please use standart library classes IPEndPoint and IPAddress? I think it's better then writing own serialization and deserialization methods

johnny-keker:
⚠️ Found out that old configs are not converted to a new format by default for now, so user gets error and proceeds with default config. This is definitely show stopper and I will fix that ASAP ⚠️

johnny-keker:
> warning Found out that old configs are not converted to a new format by default for now, so user gets error and proceeds with default config. This is definitely show stopper and I will fix that ASAP warning
> 
> Marked as [ WIP](https://github.com/vsrad/radeon-asm-tools/labels/WIP) This PR is currently work in progress until fix

Fixed in 8084564

Remove https://github.com/vsrad/radeon-asm-tools/labels/WIP, ready to review

PR #335, opened 2022-10-31
Action level lock support

kikimych:
Improved multiple user support. Replaced command level lock on the debug server to action level lock.


kikimych:
replaced with #336

PR #336, opened 2022-11-01
Action level lock

kikimych:
Improved multiple user support. Replaced command level lock on the debug server with action level lock.

### DIFF
--- a/VSRAD.DebugServer/IPC/Commands.cs
+++ b/VSRAD.DebugServer/IPC/Commands.cs
@@ -22,6 +22,11 @@ namespace VSRAD.DebugServer.IPC.Commands
         ServerNotAccepted
     }
 
+    public enum LockStatus
+    {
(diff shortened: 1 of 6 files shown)


johnny-keker:
I am facing some issues with, as far as I understand, lock releasing and re-acquiring. All details are on the screenshot below.

1. First instance of VS starts `test` action
2. Second instance of VS starts `test2` action (in less then 5 seconds after first one)

## `master` behavior (not correct)

`file.txt` is overridden by the `test2` and we see `NOT_OK` in first VS instance

## Expected behavior

`test2` is waiting when `test` is over, so we see `OK` in 1st VS instance, but file contents will be `NOT_OK` after `test2` executed

## Actual behavior

For some reason, it waits for 10 seconds no matter what and then displays `Unable to acquire lock` error

image

PR #337, opened 2022-11-08
Fixed ResizeHalf scaling with phantom column

johnny-keker:
This PR fixes `ResizeHalf` scaling mode when phantom column is visible (when data columns do not occupy the whole table)

The difference is most visible on following gif's:

## ❌ OLD / BROKEN

scale_old

## ✔️ NEW / FIXED

scale_new


Kirpich30000 (approved):
LGTM

PR #343, opened 2022-12-12
(Syntax) Make builtin function handling more accurate

timlathy:
1. Representing builtin functions with special lexer tokens is problematic because it ignores usage context (e.g. a local variable named as a builtin function is still a local variable and should be highlighted as such). With this change, builtins are treated pretty much as regular functions, except they don't show up in FL and have lower priority during name resolution.
2. The list of builtin functions is updated.


johnny-keker (approved):
LGTM, all functions highlighted, highlighted as variables if there is one

PR #346, opened 2023-02-08
Packed Int/UInt types

johnny-keker:
This PR adds new Int/UInt packed representations for Debug visualizer. One 32-bit value, two 16-bit values or four 8 bit values are supported.

image

Notes:

* this PR strongly relies on https://github.com/vsrad/radeon-asm-tools/pull/289 with some extra functionality commented out, but not removed in case we would want to bring packed representations for types other than Int/UInt
* large chunks of code that deal with computing the relative values for Slice Visualizers heatmap view were commented out either, since it need to be refactored to work with packed representation and we do not have Slice Visualizer in current upstream and the details of it's realization are still under development

### DIFF
--- a/VSRAD.Package/DebugVisualizer/ContextMenus/TypeContextMenu.cs
+++ b/VSRAD.Package/DebugVisualizer/ContextMenus/TypeContextMenu.cs
@@ -7,7 +7,7 @@ namespace VSRAD.Package.DebugVisualizer.ContextMenus
 {
     public sealed class TypeContextMenu : IContextMenu
     {
-        public delegate void TypeChanged(int rowIndex, VariableInfo type);
+        public delegate void TypeChanged(int rowIndex, VariableType type);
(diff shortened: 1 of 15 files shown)


timlathy (changes requested):
[Automated tests](https://github.com/vsrad/radeon-asm-tools/blob/flexible-datatypes-master/VSRAD.PackageTests/ProjectSystem/DebuggerIntegrationTests.cs#L43) fail to build with the the changes in this PR (`VariableType` -> `VariableInfo`). Please update the tests and check that they pass.

timlathy (commented):
The type column is too narrow to fit the new type names ("I32", "U16"):
image

(I'm not sure if you see this on your system, my setup may be different as I have display scaling set to 125%. Still, single-character types display without an issue for me, the problem is only with the new sized types.)

johnny-keker:
@timlathy ready for next round of review

PR #348, opened 2023-02-21
Copy File step refactor

kikimych:
Copy File step can now copy directories

### DIFF
--- a/RadeonAsmDebugger.sln
+++ b/RadeonAsmDebugger.sln
@@ -89,13 +89,13 @@ Global
 		{A877B375-583E-4649-9981-9D4B3F025738}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{A877B375-583E-4649-9981-9D4B3F025738}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{A877B375-583E-4649-9981-9D4B3F025738}.Release|Any CPU.Build.0 = Release|Any CPU
-		{B7222544-B13F-48C8-85E3-6BFD3D44702A}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{B7222544-B13F-48C8-85E3-6BFD3D44702A}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{B7222544-B13F-48C8-85E3-6BFD3D44702A}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{CCC917D5-5DD9-4E11-A5B3-9204E8FDC408}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
 		{CCC917D5-5DD9-4E11-A5B3-9204E8FDC408}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{CCC917D5-5DD9-4E11-A5B3-9204E8FDC408}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{CCC917D5-5DD9-4E11-A5B3-9204E8FDC408}.Release|Any CPU.Build.0 = Release|Any CPU
+		{B7222544-B13F-48C8-85E3-6BFD3D44702A}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{B7222544-B13F-48C8-85E3-6BFD3D44702A}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{B7222544-B13F-48C8-85E3-6BFD3D44702A}.Release|Any CPU.ActiveCfg = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
--- a/VSRAD.Build/VSRAD.Build.csproj
+++ b/VSRAD.Build/VSRAD.Build.csproj
@@ -23,6 +23,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <CodeAnalysisRuleSet />
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <PlatformTarget>AnyCPU</PlatformTarget>
--- a/VSRAD.DebugServer/IPC/Responses.cs
+++ b/VSRAD.DebugServer/IPC/Responses.cs
@@ -16,7 +16,10 @@ namespace VSRAD.DebugServer.IPC.Responses
         EnvironmentVariablesListed = 3,
         PutFile = 4,
         CheckOutdatedFiles = 5,
-        ListFiles = 6
+        ListFiles = 6,
+        SendFile = 7,
+        GetFile = 8,
+        PutDirectory
     }
 #pragma warning restore CA1028
 
@@ -39,6 +42,9 @@ namespace VSRAD.DebugServer.IPC.Responses
                 case ResponseType.PutFile: return PutFileResponse.Deserialize(reader);
                 case ResponseType.CheckOutdatedFiles: return CheckOutdatedFilesResponse.Deserialize(reader);
                 case ResponseType.ListFiles: return ListFilesResponse.Deserialize(reader);
+                case ResponseType.SendFile: return SendFileResponse.Deserialize(reader);
+                case ResponseType.GetFile: return GetFileResponse.Deserialize(reader);
+                case ResponseType.PutDirectory: return PutDirectoryResponse.Deserialize(reader);
             }
             throw new InvalidDataException($"Unexpected response type byte: {type}");
         }
@@ -55,6 +61,9 @@ namespace VSRAD.DebugServer.IPC.Responses
                 case PutFileResponse _: type = ResponseType.PutFile; break;
                 case CheckOutdatedFilesResponse _: type = ResponseType.CheckOutdatedFiles; break;
                 case ListFilesResponse _: type = ResponseType.ListFiles; break;
+                case SendFileResponse _: type = ResponseType.SendFile; break;
+                case GetFileResponse _: type = ResponseType.GetFile; break;
+                case PutDirectoryResponse _: type = ResponseType.PutDirectory; break;
                 default: throw new ArgumentException($"Unable to serialize {response.GetType()}");
             }
             writer.Write((byte)type);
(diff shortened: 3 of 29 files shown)


johnny-keker (changes requested):
Interface is unusable.

image

RadeonAsmDebugger.sln
```diff
@@ -89,13 +89,13 @@ Global
 		{A877B375-583E-4649-9981-9D4B3F025738}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{A877B375-583E-4649-9981-9D4B3F025738}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{A877B375-583E-4649-9981-9D4B3F025738}.Release|Any CPU.Build.0 = Release|Any CPU
-		{B7222544-B13F-48C8-85E3-6BFD3D44702A}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{B7222544-B13F-48C8-85E3-6BFD3D44702A}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{B7222544-B13F-48C8-85E3-6BFD3D44702A}.Release|Any CPU.ActiveCfg = Release|Any CPU
```
johnny-keker, line 94: Changes in this file don't make sense. We simply moved this 3 lines 4 lines further. Please revert

VSRAD.Build/VSRAD.Build.csproj
```diff
@@ -23,6 +23,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <CodeAnalysisRuleSet />
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
johnny-keker, line 26: Why do we need this property?

VSRAD.Build/VSRAD.Build.csproj
```diff
@@ -49,4 +50,4 @@ if $(ConfigurationName) == Debug echo 'Warning: you publish debug build' 1&gt;&a
   <Target Name="Clean">
     <RemoveDir Directories="$(ProjectDir)..\$(ConfigurationName)"/>
   </Target>
-</Project>
\ No newline at end of file
+</Project>
```
johnny-keker, line 53: Why do we need this line in diff?

VSRAD.DebugServer/IPC/StreamExtensions.cs
```diff
@@ -61,5 +63,139 @@ public static async Task<int> WriteSerializedMessageAsync<T>(this Stream stream,
...
+                {
+                    var sendSize = Math.Min(bytesToSend, BUFFER_SIZE);
+
+                    var readCount = reader.Read(buffer, 0, Convert.ToInt32(sendSize));
+
+                    if (readCount == 0) throw new IOException("SendFileAsync file read error");
+
+                    writer.Write(buffer, 0, readCount);
+
+                    bytesToSend -= readCount;
```
johnny-keker, line 105: Codestyle: why emptyline between every line of code?

VSRAD.DebugServer/SharedUtils/FileMetadata.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        /// <summary>
+        /// Paths use / as the directory separator. Paths ending with the slash are interpreted as directories:
+        /// this matches the behavior of zip archives used in <see cref="IPC.Commands.GetFilesCommand"/> and <see cref="IPC.Commands.PutDirectoryCommand"/>.
+        /// </summary>
+        public string relativePath_ { get; set; }
+
+        public DateTime lastWriteTimeUtc_ { get; set; }
+
+        public bool isDirectory_ { get; set; }
```
johnny-keker, line 20: Codestyle: what case is this and why is it chosen?

johnny-keker: I believe we use camel case for public properties.

VSRAD.DebugServer/SharedUtils/FileMetadata.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Xml.Serialization;
+using System.Text;
+
+namespace VSRAD.DebugServer.SharedUtils
+{
+
+    public class FileMetadata// : IEquatable<FileMetadata>
```
johnny-keker, line 10: So `IEquatable` should not be implemented?

VSRAD.DebugServer/VSRAD.DebugServer.csproj
```diff
@@ -19,11 +22,15 @@
...
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
+
   <Target Name="ValidateVersion" BeforeTargets="BeforeBuild">
     <PropertyGroup>
        <Version>$([System.Text.RegularExpressions.Regex]::Match($([System.Text.RegularExpressions.Regex]::Match($([System.IO.File]::ReadAllText("$(MSBuildProjectDirectory)\..\VSRAD.Package\source.extension.vsixmanifest")), `Version=\"[0-9,.]+\" Language=`).ToString()),`[0-9,.]+`).ToString())</Version>
     </PropertyGroup>
-    <Message Importance="High" Text="Debug server version: $(Version)"/>
+    <Message Importance="High" Text="Debug server version: $(Version)" />
```
johnny-keker, line 33: We do not need this line in diff.

johnny-keker (changes requested):

johnny-keker:
> Interface is unusable.
> 
> image

upd: fixed in [034f908](https://github.com/vsrad/radeon-asm-tools/pull/348/commits/034f908c61ec4674155e0ebeff7cdfee65ab3a45)

johnny-keker (changes requested):
Project is not building in Release configuration

image

johnny-keker:
> Project is not building in Release configuration
> 
> image

Looks like we need `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>` in both Build and DebugServer

However, jumping ahead, I don't like using unsafe code at all. As far as I can see, we can refactor code to get rid of it. But I'm still investigating it

johnny-keker:
For example, `EqualStartingCharacterCount` can be rewritten in C# way, using `Zip` from `Linq` and simple `foreach` iterator.

johnny-keker:
@kikimych please consider https://github.com/vsrad/radeon-asm-tools/commit/f93be37ae2c859b77887119a02e1d5666391ff6b

I believe this is much more laconic and simple implementation of same (am I right?) function without using unsafe code

johnny-keker (changes requested):
⚠ **Show stopper** ⚠

Current manual approach to paths handling doesn't work for Windows host -> Linux remote situation. Next image illustrates problem, Windows host is on the left, Linux remote on the right

image

Personally I am not very fond of whole `PathExtension.cs` thing, I'll try to come out with better suggestion ASAP

VSRAD.DebugServer/IPC/StreamExtensions.cs
```diff
@@ -61,5 +63,139 @@ public static async Task<int> WriteSerializedMessageAsync<T>(this Stream stream,
...
+            switch (command)
+            {
+                case SendFileCommand _:
+                    var send_command = (SendFileCommand)command;
+                    var sendPath = Path.Combine(send_command.SrcPath, send_command.Metadata.relativePath_);
+                    return await SendFileAsync(writer, sendPath, send_command.UseCompression).ConfigureAwait(false);
+                case GetFileCommand _:
+                    var receive_command = (GetFileCommand)command;
+                    var receivePath = Path.Combine(receive_command.DstPath, receive_command.Metadata.relativePath_);
+                    return await ReceiveFileAsync(writer, receivePath, receive_command.UseCompression);
```
johnny-keker, line 78: In this chunk of code `switch` already does conversion for us, however we discard it using `_`, only to do it again manually in the next line. Also, we do not use snake case anywhere in project (as far as I know)

I suggest rewriting this `switch` like this:

```c#
case SendFileCommand sf:
    var sendPath = Path.Combine(sf.SrcPath, sf.Metadata.relativePath_);
    return await SendFileAsync(writer, sendPath, sf.UseCompression).ConfigureAwait(false);
case GetFileCommand gf:
    var receivePath = Path.Combine(gf.DstPath, gf.Metadata.relativePath_);
    return await ReceiveFileAsync(writer, receivePath, gf.UseCompression);
default:
    break;
```

johnny-keker (changes requested):

VSRAD.DebugServer/SharedUtils/FileMetadata.cs
```diff
@@ -0,0 +1,59 @@
...
+        public String ToString()
+        {
+            var props = this.GetType().GetProperties();
+            var sb = new StringBuilder();
+            foreach (var p in props)
+            {
+                sb.AppendLine(p.Name + ": " + p.GetValue(this, null));
+            }
+            return sb.ToString();
+        }
```
johnny-keker, line 40: Do we really need to use reflection to iterate over 3 properties?

Also, readability can be improved. I would suggest maybe something more like

```c#
public override string ToString() => string.Join(Environment.NewLine, new string[]
{
    "Metadata = [",
    $"\tRelative Path       : {relativePath_}",
    $"\tLast Write Time UTC : {lastWriteTimeUtc_}",
    $"\tIs Directory        : {isDirectory_}",
    "]"
});
```

VSRAD.DebugServer/IPC/Responses.cs
```diff
@@ -177,6 +195,134 @@ public void Serialize(IPCWriter writer)
...
+            return string.Join(Environment.NewLine, new[] {
+                "CheckOutdatedFilesResponse",
+                $"Files = <{sb.ToString()} >"
+            });
+        }
+
+        private static XmlSerializer getFormatter()
+        {
+            return new XmlSerializer(typeof(List<FileMetadata>));
+        }
```
johnny-keker, line 313: Do we really don't have other options to serialize a List of `FileMetadata` but using `XmlSerializer`?

It seems off to me, since we never used such format before anywhere in project.

Also, this declaration clearly suggests that it was originated in some Java code, in C# we use PascalCase in such places.

johnny-keker (changes requested):

johnny-keker (changes requested):

johnny-keker:
Also, I cannot get it to work in windows `->` windows situation. As far as I understand, `dstPath` creates new directory, which is readonly, and operation cannot be completed. Also, error message in this case is not clear

image
image

johnny-keker:
When I try to copy big directory there is no descriptive message either
image
image

johnny-keker (changes requested):

PR #349, opened 2023-02-23
Add a description for token with multiple definitions

Slimakanzer:
This PR adds a description (text specified in the comment next to the token) for the token with multiple definitions such as instructions. Users don't need to click on instructions in a quick-info tooltip anymore to get a description of the instruction.

Example:
image

How it worked before:
image


Slimakanzer:
@johnny-keker ping

johnny-keker (approved):
LGTM

PR #350, opened 2023-02-24
Set new Counter minimum

johnny-keker:
This PR sets minimum value for Counter property to 1.


timlathy:
It may be a good idea to also add the new limit to the setter for `DebuggerOptions.Counter`, similarly to how it's done for `GroupSize` and `WaveSize` there. If I recall correctly, it's needed to enforce the limit on the value read from the .json config (what if it specifies `"Counter": 0`?).

PR #352, opened 2023-03-02
Fix syntax intellisense for instructions in a closure

Slimakanzer:
Previously, the quick info tooltip and the navigation didn't work with instructions in a closure like this:
```
#v_add_co_u32(v1, vcc, v2 ,v3)
```

⚠️**Low-risk change**


johnny-keker (approved):
LGTM

